PR #2, opened 2015-05-19
Feature/snakeoil certificate

feikesteenbergen:


postgres-appliance/Dockerfile
```diff
@@ -44,12 +44,12 @@ RUN curl -L https://github.com/coreos/etcd/releases/download/v${ETCDVERSION}/etc
 
 ENV PATH $PATH:/usr/lib/postgresql-${PGVERSION}/bin
 
+# Copy the snakeoil certificates for usage as dummy certificates
+RUN cp /etc/ssl/private/ssl-cert-snakeoil.key /home/postgres/dummy.key && cp /etc/ssl/certs/ssl-cert-snakeoil.pem /home/postgres/dummy.crt
```
alexeyklyukin: Why not symlink it instead?

feikesteenbergen: It is not owned by the postgres user, but is from root:ssl-cert
During runtime it cannot access the real file.

What we could do is add postgres to ssl-cert group, and not even copy it.

alexeyklyukin: You are right. In fact, I think PostgreSQL checks that the certificates are owned by postgres user and not accessible by anyone else, in the documentation there is a requirement to do chmod og-rwx server.key. So copying would be the only option.

PR #4, opened 2015-05-29
Enable pierone authentication for the etcd-cluster.

feikesteenbergen:
Extend the DEFINITION.yaml file to include the mint bucket, update the
documentation to add the dependency on mint and berry.
Expand the privileges of the IAM role to be able to get the objects out of the
specified mint bucket.

Tested and found working.

### DIFF
--- a/docs/user-guide/deploy_etcd.md
+++ b/docs/user-guide/deploy_etcd.md
@@ -7,10 +7,12 @@ The following prerequisites need to be met.
 * A [Senza Definition](http://stups.readthedocs.org/en/latest/components/senza.html#senza-definition)
 * A template can be found here: [etcd-appliance.yaml](https://github.com/zalando/spilo/blob/master/etcd-cluster-appliance/etcd-cluster.yaml)
 * The Security Group defined in the Senza Definition needs to be created
+* The appliction\_id `etcd-cluster` needs to be registered in yourturn
+* The S3 mint bucket needs to be added to the Access Control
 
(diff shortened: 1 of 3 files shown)


docs/user-guide/deploy_etcd.md
```diff
@@ -7,10 +7,12 @@ The following prerequisites need to be met.
 * A [Senza Definition](http://stups.readthedocs.org/en/latest/components/senza.html#senza-definition)
 * A template can be found here: [etcd-appliance.yaml](https://github.com/zalando/spilo/blob/master/etcd-cluster-appliance/etcd-cluster.yaml)
 * The Security Group defined in the Senza Definition needs to be created
+* The appliction\_id `etcd-cluster` needs to be registered in yourturn
+* The S3 mint bucket needs to be added to the Access Control
 
 To deploy the etcd-appliance, use the following:
 
-	senza create DEFINITION.yaml VERSION HOSTED_ZONE DOCKER_IMAGE
+	senza create DEFINITION.yaml VERSION HOSTED_ZONE DOCKER_IMAGE MINT_BUCKET
```
alexeyklyukin: Should we specify how to create mint_bucket in senza? Or provide a custom template to generate it, as well as other parameters for etcd (I think in senza one can use custom templates w/o adding them to senza itself).

feikesteenbergen: In the documentation I have updated the prerequisites to include a registration in yourturn. That should create the mint bucket as far as I know

Issue #6, opened 2015-06-04
pg_hba.conf does not allow connections from the load balancer


feikesteenbergen:
When trying to connect to a newly create cluster, my access is not allowed,

	$ ssh -f -N -L 7543:lila.acid.example.com:5432 fsteenbergen@odd.acid.example.com
	$ psql -h localhost -p 7543 -U standby -d postgres
	psql: FATAL:  no pg_hba.conf entry for host "172.31.3.222", user "standby", database "postgres", SSL on
	FATAL:  no pg_hba.conf entry for host "172.31.18.173", user "standby", database "postgres", SSL off
	$ dig +noall +answer lila.acid.example.com
	;; Warning: Message parser reports malformed message packet.
	lila.acid.example.com.	3600	IN	CNAME	internal-spilo-lila-2078097691.eu-west-1.elb.amazonaws.com.
	internal-spilo-lila-2078097691.eu-west-1.elb.amazonaws.com. 3600 IN A 172.31.15.137
	internal-spilo-lila-2078097691.eu-west-1.elb.amazonaws.com. 3600 IN A 172.31.18.173
	internal-spilo-lila-2078097691.eu-west-1.elb.amazonaws.com. 3600 IN A 172.31.3.222

PR #7, opened 2015-06-04
Allow connections to the cluster from the outside world.

feikesteenbergen:
Due to a change in Governor, pg_hba entries need to be specified to enable
connections to the cluster. This fixes https://github.com/zalando/spilo/issues/6.

### DIFF
--- a/postgres-appliance/postgres_ha.sh
+++ b/postgres-appliance/postgres_ha.sh
@@ -28,6 +28,8 @@ postgresql:
   listen: 0.0.0.0:${pg_port}
   connect_address: ${aws_private_ip}:${pg_port}
   data_dir: $PGDATA
+  pg_hba:
+  - hostssl all all 0.0.0.0/0 md5
   replication:
     username: standby
     password: standby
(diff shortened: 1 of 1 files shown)


postgres-appliance/postgres_ha.sh
```diff
@@ -28,6 +28,9 @@ postgresql:
   listen: 0.0.0.0:${pg_port}
   connect_address: ${aws_private_ip}:${pg_port}
   data_dir: $PGDATA
+  pg_hba:
+  - hostssl all all 0.0.0.0/0 md5
+  - host    all all 0.0.0.0/0 md5
```
valgog: Do you think allowing non-SSL connection is a good idea?

alexeyklyukin: It indeed has a security risk of someone running a internet-facing load balancer and no SSL, but the other side of the coin is when someone runs a server without SSL in the internal zone in VPC and will not able to connect, because there will be no entries in pg_hba.conf by default (hostssl will not be injected to the server running with SSL off). I think the second use-case will be more common in our environment.

PR #8, opened 2015-06-11
Don't start etcd proxy

CyberDem0n:
Governor could work with etcd cluster directly


postgres-appliance/postgres_ha.sh
```diff
@@ -74,20 +74,6 @@ write_postgres_yaml
...
-{
-	echo "Received TERM signal, but not doing anything"
-}
-
-# resurrect etcd if it's gone
-(
-  trap noterm SIGTERM
-  while true
-  do
-    etcd -name "proxy-$SCOPE" -proxy on  --data-dir=etcd -discovery-srv $ETCD_DISCOVERY_URL
```
alexeyklyukin: What would be the downside of leaving etcd proxy running? With it, we can always get to the container and find out the data in etcd with etcdctl without going through A-records in DNS.

CyberDem0n: Proxy could be started by Taupage.
The only what we should do - put 'etcd_discovery_domain' into TaupageConfig
Not sure that this part is already functional, because there are some hardcoded ip's and TODO's:
https://github.com/zalando-stups/taupage/blob/master/runtime/etc/init/etcd.conf https://github.com/zalando-stups/taupage/blob/master/runtime/etc/init/register-in-etcd.conf

alexeyklyukin: Sounds reasonable. I think it's ready for merge.

hjacobs:
FYI: the etcd proxy on Taupage is working, as you can see with our STUPS Hystrix Turbine dashboard

PR #11, opened 2015-07-13
Feature/exhibitor

CyberDem0n:
Write different postgres.yml depending on environment variables

It will write etcd section when ETCD_DISCOVERY_DOMAIN environment
variable is present. For zookeeper and exhibitor it checks presence of
ZOOKEEPER_HOSTS or EXHIBITOR_PORT and EXHIBITOR_HOSTS and generates
section accordingly. In case when there is no good configuration for
distributed store have been found script will exit with return code=1.

### DIFF
--- a/postgres-appliance/postgres_ha.sh
+++ b/postgres-appliance/postgres_ha.sh
@@ -32,7 +32,7 @@ __EOF__
 
     [[ -n $ZOOKEEPER_HOSTS ]] && echo "  hosts: ${ZOOKEEPER_HOSTS}" >> postgres.yml
 
-    if [[ $EXHIBITOR_HOSTS && -n $EXHIBITOR_PORT ]]; then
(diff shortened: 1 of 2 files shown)


postgres-appliance/postgres_ha.sh
```diff
@@ -12,17 +12,47 @@ function write_postgres_yaml
...
+    cat >> postgres.yml <<__EOF__
+zookeeper:
+  scope: *scope
+  session_timeout: *ttl
+  reconnect_timeout: *loop_wait
+__EOF__
+
+    [[ -n $ZOOKEEPER_HOSTS ]] && echo "  hosts: ${ZOOKEEPER_HOSTS}" >> postgres.yml
+
+    if [[ $EXHIBITOR_HOSTS && -n $EXHIBITOR_PORT ]]; then
```
alexeyklyukin: should be -n $EXHIBITOR_HOSTS

PR #12, opened 2015-07-27
Command-line client for Spilo

feikesteenbergen:
This command line client is to aid in connecting to spilo clusters running inside a vpc.
It allows to create a tunnel and connect using a single command.

The healthcheck will only connect to the governor http port of the specified host and
return its output.

The tunnel will create a tunnel for you and return a connect string.


hjacobs:
@feikesteenbergen please check the excellent "click" command line library and my utilities, I think that's nicer (esp. in the long run) for command line tools:

* https://pypi.python.org/pypi/click
* https://pypi.python.org/pypi/clickclick

Issue #31, opened 2015-11-03
Let choose default passwords via TaupageConfig


elgalu:
Right now they are [hard-coded](https://github.com/zalando/spilo/blob/2ad047c58e5eadf4a1adec833b0daa302e300894/postgres-appliance/postgres_ha.sh#L76) and changing the replica password is not straight forward, it would be nice to set them all up in the `senza.yaml` file perhaps inside TaupageConfig so it happens at spilo cluster creation time.

> if you change the password in the config file, it is not in effect immediately, it will make it into recovery.conf at some later point (status change promote/demote of replica). you will also need to change the replication role's password on the server, as the role is created only once with the initial setup of the cluster

alexeyklyukin:
I'm not sure we should put passwords in the senza.yaml. Senza templates are often committed in git, so I'd avoid putting secrets there.

What we can do instead is to provide a parameter for the senza PostgreSQL template to set the username/password. That would be the first step, but ultimately, I think we should provide a solution for Patroni (https://github.com/zalando/patroni) to change passwords at runtime. We might achieve this by storing encrypted passwords in DCS in a separate key tree and acting on changes to this tree,

elgalu:
LGTM :)

PR #35, opened 2015-12-29
Take passwords from the environment if available

feikesteenbergen:
https://github.com/zalando-stups/senza/pull/159/files


feikesteenbergen:
Addresses (part of) issue #31

Issue #23, opened 2015-09-11
Use ec2 instance ID as the name to identify a cluster


feikesteenbergen:
Currently the clustername is generated by using the HOSTNAME (which is inside Docker). This does not help in identifying which member runs on which ec2 instance.

Proposal: Include the Instance ID in the name, postfixed with a short random string to enable multiple containers running on the same ec2 instance.

PR #36, opened 2015-12-29
Rename Spilo members to match the AWS instance id.

feikesteenbergen:
For easier cross-referencing between logfiles, AWS consoles and Spilo, it is
usefull to be able to see which EC2 instance id belongs to Spilo.
The previous identifier used the Docker hostname, which does not convey
a lot of information in debugging.

To ensure replication slots can still be created, the dash needs to be transformed
into an underscore.


feikesteenbergen:
Addresses #23

PR #37, opened 2016-01-05
Initial refactoring into a tutorial with examples.

feikesteenbergen:
Refactoring the documentation, goals:

- better design documentation
- a tutorial with examples


docs/DESIGN.md
```diff
@@ -1,38 +1,19 @@
...
+# Overview
+We use PostgreSQL for this appliance because it provides a very robust RDBMS with a huge feature set.
+By using PostgreSQL we are also bound to its inherent limitations. One of those limitiations is that there can be only
+1 master, it doesn't do multi-master replication (yet). Because of this limitation we decide to build an appliance that
+has 1 master cluster and n-slave clusters.
 
-Having a single master inside this appliance means there will be a single point of failure (SPOF): the master itself. During an issue with the master the slaves will however be available for querying: it will not be a full outage. To reduce the impact of losing this SPOF we want to promote one of the slaves to a master as soon as possible once we determine the master is lost.
+Having a single master inside this appliance means there will be a single point of failure (SPOF): the master itself.
```
alexeyklyukin: I would rewrite the SPOF part:

Having a single master does not mean that the cluster will be unusable if the master fails, but makes it necessary to quickly promote the most up-to-date replica into the master role. The purpose of Patroni, the python daemon the Spilo appliance is based on, is exactly to facilitate automatic promotions of replicas to the master role in case of the former master's failure. The promotion process coverts a read-only replica into the master, which permits writes. The replicas are available for read-only queries all the time, even in the absence of the master.

I think naming the master a SPOF is too harsh: the outage of AWS is a SPOF, the non-funcitonal etcd is a SPOF, but a failed master is a temporary failure.

docs/tutorial/prerequisites.md
```diff
@@ -0,0 +1,49 @@
...
+
+```bash
+$ mai login
+Please enter password for encrypted keyring:
+Authenticating against https://aws.example.com.. OK
+Assuming role AWS Account 1234567890 (example-team): SomeRoleName.. OK
+Writing temporary AWS credentials.. OK
+```
+```bash
+$ senza list
```
alexeyklyukin: Should we do ``senza list spilo`` instead?

feikesteenbergen: It is not a prerequisite to have spilo running, that's why I left it out. It is a prerequisite to have etcd-cluster running, so I could add that?

alexeyklyukin:
+1

PR #39, opened 2016-01-25
Drop snakeoil certificate.

feikesteenbergen:
To ensure some level of reliable encryption, every instance now creates their own self-signed certificate during
boot. The previous reliance on the snakeoil certificate in the Docker image, allowed anyone (the image being public)
to decrypt ssl traffic from Spilo.
With this change we fix the issue of the certificate being known to others.


alexeyklyukin:
+1

Issue #49, opened 2016-04-04
User specified and dynamic configuration


feikesteenbergen:
The current patroni configuration is static: It does take some specific values if explicitly set (e.g. DCS) or from the environment (IP, name).

The proposal is:

1. Have some defaults
2. Override the defaults with environment specific values, examples:
    * name
    * IP
    * `shared_buffers`
3. Override these values with the user supplied values, optionally with a filter.

feikesteenbergen:
Implemented in https://github.com/zalando/patroni/tree/v1.0

PR #50, opened 2016-04-26
Pythonify startup of patroni

feikesteenbergen:
To enable better configurability of a Spilo appliance, some changes are necessary.
One of them allowing more control in overriding Patroni configuration.

Merging yamls doesn't seem to suit bash that well, therefore we now start a python script.
The backup is now a standalone bash script.

* Shared buffers is now explicitly set to 1/4 of the memory.
* Backups are scheduled using cron

Patronictl is now capable of reusing the configuration of patroni, therefore we just
link the configuration files.

Related commits:
    https://github.com/zalando-stups/senza/commit/3b548db6d976ca224e629dcabcfa07fba175d400


postgres-appliance/Dockerfile
```diff
@@ -53,30 +53,28 @@ WORKDIR $PGHOME
 RUN pip install patroni==$PATRONIVERSION
 
 # WAL-e requests a relatively new version of requests, install other modules via pip as well
-RUN pip install requests wal-e --upgrade
+RUN pip install requests wal-e --upgrade pystache
```
a1exsh: Should we move `--upgrade` before the first package name?  It would be easier to understand, IMO.

postgres-appliance/Dockerfile
```diff
@@ -53,30 +53,28 @@ WORKDIR $PGHOME
...
-# Shoud be 3am by GMT+2 aka Berlin Summer Time
-ENV BACKUP_HOUR 1
-ENV WALE_BACKUP_THRESHOLD_MEGABYTES 1024
-ENV WALE_BACKUP_THRESHOLD_PERCENTAGE 30
-ENV WALE_ENV_DIR $PGHOME/etc/wal-e.d/env
+
 ENV LC_ALL en_US.utf-8
+
+# Allow cron to be started by postgres
+RUN echo "postgres ALL=(ALL:ALL) NOPASSWD: /usr/sbin/cron" > /etc/sudoers #Allow cron to be started by postgres
```
a1exsh: Duplicated comment...

postgres-appliance/postgres_backup.sh
```diff
@@ -0,0 +1,44 @@
...
+        sleep 5
+    done
+
+    echo "ERROR: cluster was not available after $SECONDS seconds" >&2
+    return 1
+}
+
+[[ -z $2 ]] && echo "Usage: $0 WALE_ENV_DIR PGDATA" && exit 1
+
+if [[ ! -z ${INITIAL_BACKUP+1} ]]
```
alexeyklyukin: why  +1 ?

postgres-appliance/postgres_ha.py
```diff
@@ -0,0 +1,286 @@
...
+    For anything else, the value of a is returned"""
+
+    if type(a) is dict and type(b) is dict:
+        for key in b:
+            if key in a:
+                a[key] = deep_update(a[key], b[key])
+            else:
+                a[key] = b[key]
+        return a
+    if type(a) is list and type(b) is list:
```
alexeyklyukin: I think your list merging might be a little more sophisticated, i.e. to avoid duplicates (note though, that doing it via list(set)) is a bad idea, since in pg_hba case you should care about the order of records. Something like [a + [i in b if i not in a]]

postgres-appliance/postgres_ha.py
```diff
@@ -0,0 +1,286 @@
...
+    restore_command: envdir "{{WALE_ENV_DIR}}" wal-e --aws-instance-profile wal-fetch "%f" "%p" -p 1
+'''
+
+
+def get_instance_meta_data(key):
+    try:
+        result = requests.get('http://instance-data/latest/meta-data/{}'.format(key), timeout=2)
+        if result.status_code != 200:
+            raise Exception('Received status code {} ({})'.format(result.status_code, result.text))
+        return requests.get('http://instance-data/latest/meta-data/{}'.format(key), timeout=2).text
```
alexeyklyukin: why not result.text?

postgres-appliance/postgres_ha.py
```diff
@@ -0,0 +1,286 @@
...
+    configure_patronictl(patroni_configfile, os.path.join(placeholders['PGHOME'], '.config', 'patroni',
+                         'patronictl.yaml'))
+
+    # WAL-E
+    write_wale_command_environment(placeholders)
+
+    # # We run cron, to schedule recurring tasks on the node
+    logging.info('Starting up cron')
+    subprocess.call(['/usr/bin/sudo', '/usr/sbin/cron'])
+    write_crontab(placeholders, os.environ.get('PATH'))
```
alexeyklyukin: Would it make more sense to first write the crontab, and only afterwards launch the cron daemon?

postgres-appliance/postgres_ha.py
feikesteenbergen: Yes, should be result.text

postgres-appliance/postgres_ha.py
feikesteenbergen: We could do that, I don't easily see where we would have a legitimate list with duplicates.

postgres-appliance/postgres_backup.sh
feikesteenbergen: Variable expansion, I just want to know whether the variable is set or not, I don't care about the value, see:
http://stackoverflow.com/questions/3601515/how-to-check-if-a-variable-is-set-in-bash

```bash
$ export TEST=
$ [[ ! -z ${TEST} ]] && echo "Yes"
$ [[ ! -z ${TEST+1} ]] && echo "Yes"
Yes
$ unset TEST
$ [[ ! -z ${TEST+1} ]] && echo "Yes"
$
```

alexeyklyukin: ok, although I find the syntax ugly (and you always pass a value to INITIAL_BACKUP).

postgres-appliance/postgres_ha.py
alexeyklyukin: What about merging pg_hba lines?

alexeyklyukin:
👍

postgres-appliance/postgres_backup.sh
a1exsh: What about good old "x$VAR != x" -- think this is portable?

Side note: line notes made from commit page don't show up in PR...

postgres-appliance/postgres_backup.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
```
a1exsh: Dare I say that this should be a separate commit?

feikesteenbergen:
fixes #49 

Tests succeeded:

- shared_buffers is set to 1/4 of the available memory

**t2.micro**
```bash
postgres=# show shared_buffers ;
 shared_buffers
----------------
 248MB
(1 row)
```

**t2.medium**
```bash
postgres=# show shared_buffers ;
 shared_buffers
----------------
 987MB
(1 row)
```

- parameters set using PATRONI_CONFIGURATION are overridden

```bash
postgres@6b33acfdfa64:~$ echo "${PATRONI_CONFIGURATION}"
postgresql:
  parameters:
    max_connections: 123
#   pg_hba:
#   - hostnossl all all all reject
#   - hostssl   all all all md5

postgres@6b33acfdfa64:~$ psql -AtX --command="show max_connections"
123
```

- backup is initiated as part of container start
- backup is initiated by cron
- To be determined: migrate a spilo-9.5:0.80-p1 (using `postgres_ha.sh`) to a spilo-9.5:0.80-SNAPSHOT (using `postgres_ha.py`) without issue

postgres-appliance/postgres_backup.sh
feikesteenbergen: Different commit: could be, I'll have to squash some commits then, different PR: No, as the previous script (postgres_ha.sh) is split into postgres_ha.py and postgres_backup.sh (and cron).

PR #53, opened 2016-04-28
Add scm-source to Dockerfile when building.

feikesteenbergen:
To be compliant with Stups we need to ensure we add an up-to-date
scm-source.json file to the root of the Docker container.

### DIFF
--- a/postgres-appliance/build.sh
+++ b/postgres-appliance/build.sh
@@ -20,7 +20,7 @@ cat > scm-source.json <<__EOT__
     "url": "$URL",
     "revision": "$REV",
     "author": "$GITAUTHOR",
-    "status": "'$STATUS'"
+    "status": "$STATUS"
 }
 __EOT__
 
--- a/postgres-appliance/build.sh
+++ b/postgres-appliance/build.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+
+DOCKERCMD="docker build"
+
+function usage()
+{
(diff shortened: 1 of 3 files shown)


postgres-appliance/build.sh
```diff
@@ -0,0 +1,34 @@
...
+URL=$(git config --get remote.origin.url)
+STATUS=$(git status --porcelain)
+GITAUTHOR=$(git show -s --format="%aN <%aE>" "$REV")
+
+cat > scm-source.json <<__EOT__
+{
+    "url": "$URL",
+    "revision": "$REV",
+    "author": "$GITAUTHOR",
+    "status": "'$STATUS'"
```
hjacobs: I see additional single quotes here, compare with http://docs.stups.io/en/latest/user-guide/application-development.html#scm-source-json

feikesteenbergen: They seem unessesary, yes, however the same page also has the version with quotes:

https://github.com/zalando-stups/documentation/blame/master/user-guide/application-development.rst#L72

Will remove from this PR

hjacobs: @feikesteenbergen thanks, will fix it in the docs..

hjacobs: @feikesteenbergen no, the docs are right (single quotes are needed to end the outer string and to interpret the env var)

alexeyklyukin:
👍

PR #57, opened 2016-04-29
Conditionally write Patroni configuration to disk

feikesteenbergen:
Another layer of defense; by not writing the actual configuration to disk, we have one less
place where the plaintext passwords are stored.
Patronictl still requires the configurationfile, therefore we only write the necessary keys,
again we don't write the actual passwords to disk.

Usecase: A postgres superuser can read the patroni configuration. If it is not written,
it cannot be examined. Neither can the environment of patroni be examined by another process of postgres.


alexeyklyukin:
👍

PR #58, opened 2016-05-04
Limit the maximum amount of connections

feikesteenbergen:
Although adding more allowed connections on bigger instances, it
doesn't make sense to not have an upper limit for this autoscaling.
Users are still free to override max_connections if they need to,
we just make sure we stay within a relatively sane range.


alexeyklyukin:
👍

PR #59, opened 2016-05-06
Feature/reduce docker size

CyberDem0n:


hjacobs:
Hint: you can use "docker-squash" to minimize the final Docker image size

feikesteenbergen:
We don't have any unnecessary stuff installed. 
Any temporary stuff is done in one Docker command (install, compile, remove, etc)

Docker squash does help: it reduces by 6MB.

feikesteenbergen:
+1

PR #60, opened 2016-05-20
Feature/improve spilo logging

alexeyklyukin:
- Set log line prefix to include database, host, application,
  session and transaction IDs, number of log line for the session
  and the process ID.
- Enable logging of checkpoints, statements longer than 0.5s, lock
  waits, autovacuum runs, connections and disconnections, DDL and
  temp files.


feikesteenbergen:
👍

PR #61, opened 2016-05-20
Migrate to supervisord

feikesteenbergen:
As the Spilo container assumes more and more roles it
starts to be cumbersome to manage all the processes inside the container.

supervisord allows us to run multiple processes, restarting if necessary
and allowing future improvements (stunnel for ldap, ticker daemon for pgq)
to be more easily implemented.

### DIFF
--- a/postgres-appliance/Dockerfile
+++ b/postgres-appliance/Dockerfile
@@ -8,7 +8,7 @@ RUN export DEBIAN_FRONTEND=noninteractive \
     && echo 'APT::Install-Suggests "0";' >> /etc/apt/apt.conf.d/01norecommend \
 
     && apt-get upgrade -y \
-    && apt-get install -y curl ca-certificates jq pv vim gdb strace \
+    && apt-get install -y curl ca-certificates jq pv vim gdb strace supervisor \
 
     ## Make sure we have a en_US.UTF-8 locale available
     && localedef -i en_US -c -f UTF-8 -A /usr/share/locale/locale.alias en_US.UTF-8 \
@@ -62,6 +62,7 @@ RUN export DEBIAN_FRONTEND=noninteractive \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/* /root/.cache
 
+
 # install etcdctl
 ENV ETCDVERSION 2.3.3
 RUN curl -L https://github.com/coreos/etcd/releases/download/v${ETCDVERSION}/etcd-v${ETCDVERSION}-linux-amd64.tar.gz \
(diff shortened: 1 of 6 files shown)


postgres-appliance/Dockerfile
```diff
@@ -62,27 +62,33 @@ RUN export DEBIAN_FRONTEND=noninteractive \
...
 
+
 # install etcdctl
 ENV ETCDVERSION 2.3.3
 RUN curl -L https://github.com/coreos/etcd/releases/download/v${ETCDVERSION}/etcd-v${ETCDVERSION}-linux-amd64.tar.gz \
     | tar xz -C /bin --strip=1 --wildcards --no-anchored etcdctl
 
+# install supervisord
+RUN    apt-get update \
+    && apt-get upgrade -y \
```
alexeyklyukin: Why do we need apt-get update? That may install quite a lot of things we don't really need.

CyberDem0n:
Will it propagate SIGTERM to child processes? It's important for Patroni.

feikesteenbergen:
Yes, it propagates SIGTERM it to the child processes:

http://supervisord.org/running.html#signal-handlers

We can even control the behaviour and timeouts quite easily:
http://supervisord.org/configuration.html#program-x-section-values

postgres-appliance/Dockerfile
```diff
@@ -62,27 +62,28 @@ RUN export DEBIAN_FRONTEND=noninteractive \
...
+ENV WALE_ENV_DIR=$PGHOME/etc/wal-e.d/env
 
 # Set PGHOME as a login directory for the PostgreSQL user.
 RUN usermod -d $PGHOME -m postgres
 
-# Allow cron to be started by postgres
-RUN echo "postgres ALL=(ALL:ALL) NOPASSWD: /usr/sbin/cron" > /etc/sudoers
-
-ADD scm-source.json postgres_ha.py postgres_backup.sh /
+ADD configure_spilo.py launch.sh postgres_backup.sh /
```
alexeyklyukin: I think we deleted scm-source.json, but never added it.

postgres-appliance/configure_spilo.py
```diff
@@ -42,7 +66,9 @@ def write_certificates(environment):
         output, _ = p.communicate()
         logging.debug(output)
 
-    os.chmod(environment['SSL_PRIVATE_KEY_FILE'], 0o600)
+    gid = os.stat(environment['PGHOME']).st_gid
+    os.chmod(environment['SSL_PRIVATE_KEY_FILE'], 0o640)
+    os.chown(environment['SSL_PRIVATE_KEY_FILE'], -1, gid)
```
alexeyklyukin: Why did you decide to adjust the group and not the owner of the private key file?

postgres-appliance/configure_spilo.py
```diff
@@ -229,49 +257,45 @@ def get_dcs_config(config, placeholders):
...
+               os.path.join(placeholders['WALE_ENV_DIR'], 'WALE_S3_ENDPOINT'), overwrite)
+
 
-    with open(os.path.join(placeholders['WALE_ENV_DIR'], 'WALE_S3_ENDPOINT'), 'w') as f:
-        f.write('https+path://s3-{}.amazonaws.com:443'.format(az[:-1]))
+def write_crontab(placeholders, path, overwrite):
 
+    if not overwrite:
+        with open(os.devnull, 'w') as devnull:
+            cron_exit = subprocess.call(['sudo', '-u', 'postgres', '-l'], stdout=devnull, stderr=devnull)
```
alexeyklyukin: sudo -u postgres -l doesn't have anything to do with the crontab. Did you intend to write sudo -u postres crontab -l?

postgres-appliance/configure_spilo.py
feikesteenbergen: I thought it would prevent the postgres os-user from changing the file; but that assumption is wrong, as the directory is owned by postgres.
Switching to uid makes more sense.

alexeyklyukin:
+1

PR #64, opened 2016-05-27
Add pgq into Spilo

feikesteenbergen:
Some teams require pgq to run. After some analysis it shows pgq
can be added without a lot of configuration. Adding the pgq
extension will be picked up by pgq within a few minutes.

To reduce logging noise, we do prepend it with patroni_wait


alexeyklyukin:
+1

slitsche:
👍

PR #66, opened 2016-05-31
Introduce an stunnel to connect to an ldaps

feikesteenbergen:
This feature allows us to connect to an ldap server that only
allows a direct SSL connection. This however is not supported for
PostgreSQL, see [1]. By adding an stunnel process if and only
if an LDAP_URL is specified, we ensure only those Spilo's that
require this stunnel will have it configured.

A dynamic configuration would be nice, currently it is static due
to the Certificate Chain having only specific certificates.
Any future dynamic configuration needs to setup a secure way
of setting these certificates.

[1] https://www.postgresql.org/docs/current/static/auth-methods.html#AUTH-LDAP


alexeyklyukin:
+1

postgres-appliance/configure_spilo.py
```diff
@@ -287,6 +290,47 @@ def write_crontab(placeholders, path, overwrite):
     c.communicate(input='\n'.join(lines).encode())
 
 
+def write_ldap_configuration(placeholders, overwrite):
+    if not 'LDAP_URL' in placeholders:
```
a1exsh: The idiom is `key not in dict`, I believe ;-)  Also, this could be written as
```
ldap_url = placeholders.get('LDAP_URL')
if ldap_url is NoneL
    # complain and bail out
```

postgres-appliance/configure_spilo.py
```diff
@@ -287,6 +290,47 @@ def write_crontab(placeholders, path, overwrite):
...
+
+    r = urlparse(placeholders['LDAP_URL'])
+    if not r.scheme:
+        logging.error('LDAP_URL should contain a scheme')
+        logging.info(r)
+        return
+
+    host, port = r.hostname, r.port
+    if not port:
+        port = 636 if r.scheme == 'ldaps' else '389'
```
a1exsh: `type is int else str`

PR #68, opened 2016-06-01
Include some extensions from PGXN

feikesteenbergen:
Some extensions were requested that are easily injected, as they don't have
non-postgresql dependencies.


avaczi:
+1

alexeyklyukin:
+1

PR #69, opened 2016-06-02
Ensure ownership of PGDATA

feikesteenbergen:
When restoring from an EBS Snapshot, it may happen that the PGDATA
directory (and files) is owned by another user. This fixes that.


a1exsh:
Nice catch :+1:

alexeyklyukin:
+1

PR #70, opened 2016-06-15
Include most pls available from the repositories

feikesteenbergen:
The added pl's don't add a lot of bytes, however, they allow
a very wide (or even strange) variety of pl's to be used.


alexeyklyukin:
+1

PR #73, opened 2016-06-28
Enable possibility from restore from wal-e

### LINKED ISSUE #72: Restore with wal_e doesn't work
Missing:
```yaml
postgresql:
  create_replica_method:
  - wal_e
  - basebackup
```

So it always defaults to `basebackup` method

feikesteenbergen:
This possibility was accidentally removed in commit
55fd3aa121e2ded2654a2f4d4c4ff96bad061f40

Fixes #72

### DIFF
--- a/postgres-appliance/configure_spilo.py
+++ b/postgres-appliance/configure_spilo.py
@@ -131,6 +131,7 @@ postgresql:
   create_replica_method:
     - wal_e
     - basebackup
+  maximum_lag_on_failover: 33554432
   wal_e:
     command: patroni_wale_restore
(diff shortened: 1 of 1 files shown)


alexeyklyukin:
We also need to set the no_master: 1 for the wal-e, otherwise, it won't be considered when the master is not there.

feikesteenbergen:
Would be something like this, right?:

```yaml
create_replica_method:
  - wal_e:
      no_master: 1
  - basebackup

alexeyklyukin:
Nope, it should be a separate section under postgresql. It's already there actually.

alexeyklyukin:
+1

PR #85, opened 2016-08-25
Re-enable dynamic tuning of shared_buffers

feikesteenbergen:
In commit c74a62f4fcc448069f5e9a579322ea6deec1bdc5 the configuration was changed to a Patroni 1.0 compatible
configuration. However, shared_buffers was pinned to 500MB with this.

### DIFF
--- a/postgres-appliance/configure_spilo.py
+++ b/postgres-appliance/configure_spilo.py
@@ -111,7 +111,7 @@ bootstrap:
         wal_keep_segments: 8
         wal_log_hints: 'on'
         max_wal_senders: 5
-        shared_buffers: 500MB
+        shared_buffers: {{postgresql.parameters.shared_buffers}}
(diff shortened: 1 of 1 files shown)


feikesteenbergen:
👍

postgres-appliance/configure_spilo.py
```diff
@@ -111,7 +111,7 @@ def deep_update(a, b):
         wal_keep_segments: 8
         wal_log_hints: 'on'
         max_wal_senders: 5
-        shared_buffers: 500MB
+        shared_buffers: {{postgresql.parameters.shared_buffers}}
```
CyberDem0n: I think `shared_buffers` should not be configured cluster-wide, you need to move it to 'postgresql.parameters'

feikesteenbergen:
👍

CyberDem0n:
:+1:

PR #88, opened 2016-09-02
Use the bucket region for WAL-E Endpoint

feikesteenbergen:
When doing cross-region replication, it may occur that WAL-E requests s3 objects
from a different region. By default, WAL-E is configured to connect to the endpoint
of the region where it is running.
This patch changes the logic so that we choose the endpoint based upon the name
of the bucket. If that fails, we use the previous behaviour.


feikesteenbergen:
👍

feikesteenbergen:
Did a deployment, which didn't cause issues

feikesteenbergen:
👍

CyberDem0n:
:+1:

PR #89, opened 2016-09-02
Spin up local etcd if DEVELOP is specified

feikesteenbergen:
Multiple developers have asked for a Spilo like environment including pgq,
postgis etc.
By starting up etcd if DEVELOP is specified, we can easily ask people to deploy against
what we actually run in production.


alexeyklyukin:
👍

feikesteenbergen:
👍

PR #91, opened 2016-09-08
Create a directory for storing logs of PostgreSQL

feikesteenbergen:
Some Spilo's currently log to the filesystem inside docker. If these
files are stored inside $PGDATA, they are also part of any basebackup or WAL-E backup.
When these files grow big, WAL-E fails to backup, because it assumes no files inside $PGDATA
are bigger than 1.5GB.

There now is a pull request to address this at the WAL-E level: https://github.com/wal-e/wal-e/pull/278
however, it seems usefull to allow logs to persist outside the $PGDATA directory.

This only creates the directory, it does not configure PostgreSQL to actually use this directory, that task
is for the senza template.


CyberDem0n:
:+1:

feikesteenbergen:
👍

PR #94, opened 2016-09-15
Ensure the logfiles are owned by postgres

feikesteenbergen:
If PostgreSQL cannot write to the csv logs it will flood its diagnostic log.
This has already happened and can write > 100GB within a night.


avaczi:
:+1:

feikesteenbergen:
👍

PR #95, opened 2016-09-21
allow to configure the kubernetes namespace

linki:
this allows to inject a namespace to generate the url for the pod.

currently the namespace is hard-coded which makes it impossible to run patroni in another namespace. the namespace can be configured with an env var called `POD_NAMESPACE` and will fall back to `default` if it isn't set.

`POD_NAMESPACE` is the canonical env var name for injecting the namespace into a pod.


alexeyklyukin:
👍

linki:
👍

PR #97, opened 2016-10-13
Use PROVIDER_LOCAL when DEVELOP env is set

CyberDem0n:
It doesn't make much sense determining provider in this case.
Also it will make it possible to run spilo as standalone app on AWS with
for example jenkins.


avaczi:
:+1:

avaczi:
:+1:

CyberDem0n:
:+1:

PR #99, opened 2016-10-27
use spilo with local

andre82hh:
Hi,

i try to use spilo with local minikube. But the leader election do not work for me. So i had to add PROVIDER_LOCAL while configure.

### DIFF
--- a/postgres-appliance/configure_spilo.py
+++ b/postgres-appliance/configure_spilo.py
@@ -257,7 +257,7 @@ def get_placeholders(provider):
     placeholders.setdefault('WALE_BACKUP_THRESHOLD_PERCENTAGE', 30)
     placeholders.setdefault('WALE_ENV_DIR', os.path.join(placeholders['PGHOME'], 'etc', 'wal-e.d', 'env'))
 
-    if provider in (PROVIDER_AWS, PROVIDER_GOOGLE):
+    if provider in (PROVIDER_AWS, PROVIDER_GOOGLE, PROVIDER_LOCAL):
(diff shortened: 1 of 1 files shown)


postgres-appliance/configure_spilo.py
```diff
@@ -257,7 +257,7 @@ def get_placeholders(provider):
     placeholders.setdefault('WALE_BACKUP_THRESHOLD_PERCENTAGE', 30)
     placeholders.setdefault('WALE_ENV_DIR', os.path.join(placeholders['PGHOME'], 'etc', 'wal-e.d', 'env'))
 
-    if provider in (PROVIDER_AWS, PROVIDER_GOOGLE):
+    if provider in (PROVIDER_AWS, PROVIDER_GOOGLE, PROVIDER_LOCAL):
```
alexeyklyukin: We don't really want placeholder.setdefault('USE_WALE', True) for the PROVIDER_LOCAL, since WAL-E is not able when you run in minukube. I suggest instead to move USE_K8S after the block in lines 260: 275, which should set a callback script for K8S not depending on a current provider.

andre82hh: thx for advise! I've changed

alexeyklyukin:
👍

LappleApple:
👍

LappleApple:
@alexeyklyukin @CyberDem0n Guys, this needed one more Zappr approval. I made it, so if you are still willing to merge I guess go ahead and do so.

CyberDem0n:
PR is obsolete.
Spilo already can work everywhere.

PR #103, opened 2016-11-07
Configuration can be set with $SPILO_CONFIGURATION.

### LINKED ISSUE #102: Spilo reuses PATRONI_CONFIGURATION in order to supply custom configuration
I'm thinking about making some changes that may potentially break things for the already existing clusters: the variable PATRONI_CONFIGURATION that is set by the senza base template and used in Spilo in order to merge the default configuration with the custom one, happens to also be used by Patroni in order to configure itself entirely from the environment variables. So far, it doesn't break things for Patroni, because the environment is cleaned up at the point of executing Patroni from Spilo, but it does lead to issues when someone tries to execute patronictl inside the docker container under the default (root) user, since patronictl observes the partial  PATRONI_CONFIGURATION set by Spilo and interprets it as a complete Patroni configuration, resulting in an exception, since some required parameters (namely, authentication related) are not set.

Therefore, in order to avoid breaking existing setups I will introduce the SPILO_CONFIGURATION variable, that will be examined alongside PATRONI_CONFIGURATION, and PATRONI_CONFIGURATION will be ignored if SPILO_CONFIGURATION is set.

alexeyklyukin:
Previously, Spilo used PATRONI_CONFIGURATION in order to read the
non-default configuration for Patroni (the default one is hard-coded
in the configure_spilo.py. Because the envionment variable with the
same name is also used by Patroni for a completely different purpose,
it lead to a conflict that was hard to debug. Therefore, we need to
rename it, but in order to keep the compatibility with the already
deployed clusters, both the old and the new (SPILO_CONFIGURATION)
names will be accepted by the Docker image with the preference for
the new one if both are set. That will allow us to change the
deployment configuration to use the new name without breaking already
deployed clusters in the case we will need to update their configuration
on the fly.

This should fix #102

### DIFF
--- a/postgres-appliance/configure_spilo.py
+++ b/postgres-appliance/configure_spilo.py
@@ -333,7 +333,7 @@ etcd:
         config = {'etcd': defaults['etcd']}
         config['etcd']['discovery_srv'] = placeholders['ETCD_DISCOVERY_DOMAIN']
     else:
-        pass  # Configuration can also be specified using PATRONI_CONFIGURATION
+        pass  # Configuration can also be specified using either SPILO_CONFIGURATION or PATRONI_CONFIGURATION
 
     return config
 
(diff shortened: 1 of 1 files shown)


a1exsh:
:+1:

alexeyklyukin:
👍

PR #105, opened 2016-11-18
Make spilo more k8s-compatible

CyberDem0n:


postgres-appliance/callback_role.py
```diff
@@ -8,57 +8,68 @@
...
+    pod_namespace = os.environ.get('POD_NAMESPACE', read_first_line(KUBE_NAMESPACE_FILENAME)) or 'default'
+    api_url = KUBE_API_URL.format(pod_namespace, os.environ['HOSTNAME'])
+    body = json.dumps({'metadata': {'labels': {LABEL: new_role}}})
     for i in range(NUM_ATTEMPTS):
         try:
-            r = requests.patch(url, headers=headers, data=json.dumps(data), verify=CA_CERT_FILENAME)
+            token = read_token()
+            if not token:
+                logger.warning('Unable to read K8S authorization token')
+                continue
```
a1exsh: Should there be no delay between failed attempts to read the file?  I'd think if it fails, then trying 10 times w/o a delay will fail for sure as well.

postgres-appliance/spilok8s.yaml
```diff
@@ -1,41 +1,25 @@
...
-        app: spilodemo
+        app: spilo
+        spilo-cluster: *cluster_name
       annotations:
         pod.alpha.kubernetes.io/initialized: "true"
     spec:
       containers:
-      - env:
-        - name: PGPASSWORD_SUPERUSER
-          value: tea
```
lotharschulz: I know ```tea```is a place holder, but we should leverage the secrets [concept kubernetes](http://kubernetes.io/docs/user-guide/secrets/) provides. Worked for quite well: https://github.com/lotharschulz/akkahttp-playground/blob/master/pod-config.yaml#L14

a1exsh:
Should the name of file `postgres-appliance/callback_role.py` reflect that it's k8s-specific?

postgres-appliance/spilok8s.yaml
```diff
@@ -44,16 +28,44 @@ spec:
         volumeMounts:
         - mountPath: /home/postgres/pgdata
           name: pgdata
-        - mountPath: /etc/wal-e
-          name: wale-config
-          readOnly: true
+        env:
+        - name: ETCD_HOST
+          value: 'etcd.default.svc.cluster.local:2379' # where is your etcd?
```
lotharschulz: Services like ETCD should run as [kubernetes services](http://kubernetes.io/docs/user-guide/services/) correct? If so, could we leverage [label selectors](http://kubernetes.io/docs/user-guide/labels/#label-selectors) to connect to the ETCD cluster?

postgres-appliance/spilok8s.yaml
CyberDem0n: This line was actually removed and  `tea` was not a placeholder, but the real password :)

postgres-appliance/configure_spilo.py
```diff
@@ -256,26 +261,29 @@ def get_placeholders(provider):
...
+
         # Kubernetes requires a callback to change the labels in order to point to the new master
         if USE_K8S:
             placeholders.setdefault('CALLBACK_SCRIPT', '/callback_role.py')
         elif provider == PROVIDER_AWS:  # AWS specific callback to tag the instances with roles
             placeholders.setdefault('CALLBACK_SCRIPT', 'patroni_aws')
 
     else:  # avoid setting WAL-E archive command and callback script for unknown providers (i.e local docker)
-        placeholders.setdefault('USE_WALE', False)
         placeholders.setdefault('CALLBACK_SCRIPT', '')
```
alexeyklyukin: For K8S, the callback should be set to /callback_role no matter which provider we are using (it can be neither Google nor AWS in case of a minikube).

postgres-appliance/spilok8s.yaml
lotharschulz: So you want to go w/o services and labels for this case?

postgres-appliance/spilok8s.yaml
```diff
@@ -69,23 +81,30 @@ spec:
...
     spilo-role: master
 
 ---
 apiVersion: v1
 kind: Secret
 metadata:
-  name: wale-secret
+  name: &cluster_name spilodemo
+  labels:
+    app: spilo
```
alexeyklyukin: We don't use those labels anywhere, so why do we need them?

postgres-appliance/configure_spilo.py
CyberDem0n: Nice catch.

postgres-appliance/spilok8s.yaml
CyberDem0n: I think they are still useful. Thus we can easily find all objects related to all spilos.

postgres-appliance/spilok8s.yaml
CyberDem0n: Yep. It is simpler for everybody.

a1exsh:
:+1:

postgres-appliance/callback_role.py
```diff
@@ -8,57 +8,67 @@
...
+                r = requests.patch(api_url, data=body, verify=KUBE_CA_CERT,
+                                   headers={'Content-Type': 'application/strategic-merge-patch+json',
+                                            'Authorization': 'Bearer {0}'.format(token)})
+                if r.status_code >= 300:
+                    logger.warning('Unable to change the %s label to %s: %s', LABEL, new_role, r.text)
+                else:
+                    break
             else:
-                break
+                logger.warning('Unable to read K8S authorization token')
```
hjacobs: I would avoid the "K8s" abbrevation and use "Kubernetes" everywhere.

Raffo: +1 on that.

postgres-appliance/configure_spilo.py
```diff
@@ -219,14 +217,21 @@ def get_instance_metadata(provider):
                 'id': socket.gethostname(),
                 'zone': 'local'}
 
+    if USE_K8S:
```
Raffo: For naming consistency, you have `KUBE` above, `K8S` here, maybe we should be somehow consistent. Do you think that `KUBERNETES` would be too long?

hjacobs: Yes, please use "Kubernetes" everywhere for consistency and reader convenience, i.e. `USE_KUBERNETES` here.

postgres-appliance/spilok8s.yaml
```diff
@@ -1,41 +1,25 @@
...
 spec:
   replicas: 3
-  serviceName: spilodemo
+  serviceName: *cluster_name
   template:
     metadata:
       labels:
-        app: spilodemo
+        app: spilo
+        spilo-cluster: *cluster_name
```
Raffo: Should we use camel case like for the Kubernetes defined parts (i.e. `serviceName`).

hjacobs:
@alexeyklyukin thanks.

hjacobs:
So how do I use the YAML file? I tried:

```
zalando-kubectl apply -f spilok8s.yaml
```

This worked, but now how to use it?

alexeyklyukin:
> On 21 Nov 2016, at 14:44, Henning Jacobs <notifications@github.com> wrote:
> 
> So how do I use the YAML file? I tried:
> 
> zalando-kubectl apply -f spilok8s.yaml
> This worked, but now how to use it?
> 
> 

You have to supply some parameters of the cluster manually in the manifest, the most important one is the cluster_name. This week we are going to write the documentation on how to deploy Spilo based on this manifest in the hackweek account.

Kind regards,
--
Oleksii Kliukin | 5432 Engineer | phone: +4917612759799 |  hipchat: @OleksiiKliukin

alexeyklyukin:
👍

CyberDem0n:
:+1:

PR #106, opened 2016-11-22
Fix Endpoints labels

CyberDem0n:


postgres-appliance/spilok8s.yaml
```diff
@@ -85,7 +85,8 @@ kind: Endpoints
 metadata:
   name: &cluster_name spilodemo
   labels:
-    app: *cluster_name
+    application: spilo:
+    spilo-cluster: *cluster_name
```
a1exsh: So we now have camelCase, snake_case, and now lisp-case or what? ;-)

postgres-appliance/spilok8s.yaml
```diff
@@ -85,7 +85,8 @@ kind: Endpoints
 metadata:
   name: &cluster_name spilodemo
   labels:
-    app: *cluster_name
+    application: spilo:
```
ikitiki: Should there be a colon in the end?

CyberDem0n: Definitely not, thank you, fixed

postgres-appliance/spilok8s.yaml
```diff
@@ -85,7 +85,8 @@ kind: Endpoints
 metadata:
   name: &cluster_name spilodemo
   labels:
-    app: *cluster_name
+    application: spilo
+    spilo-cluster: *cluster_name
```
a1exsh: Why dash in `spilo-cluster`?  We already have camelCase and snake_case there...

CyberDem0n: Because it is already 'spilo-cluster' in all other places. Moreover, we have 'spilo-role' label.

a1exsh: That's not very consistent.

alexeyklyukin:
👍

postgres-appliance/spilok8s.yaml
CyberDem0n: Not consistent with what? camelCase is used for kubernetes objects and entities, we are using dashes for our objects (spilo-cluster, spilo-role).
And `cluster_name` is a very special case, it is used and the name for a yaml anchor and alias.

CyberDem0n:
:+1:

PR #108, opened 2016-12-08
Set different TMPDIR for wal-e and install pg_view

CyberDem0n:
TMPDIR for wal-e will be located on separate volume.
In addition to that export necessary env variables to make pg_view and
patronictl work when you run them with docker exec

### DIFF
--- a/postgres-appliance/configure_spilo.py
+++ b/postgres-appliance/configure_spilo.py
@@ -251,6 +251,7 @@ def get_placeholders(provider):
     placeholders.setdefault('BACKUP_SCHEDULE', '00 01 * * *')
     placeholders.setdefault('CRONTAB', '[]')
     placeholders.setdefault('PGROOT', os.path.join(placeholders['PGHOME'], 'pgroot'))
+    placeholders.setdefault('WALE_TMPDIR', os.path.abspath(os.path.join(placeholders['PGROOT'], '../tmp')))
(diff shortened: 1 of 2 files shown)


postgres-appliance/configure_spilo.py
```diff
@@ -251,6 +251,7 @@ def get_placeholders(provider):
     placeholders.setdefault('BACKUP_SCHEDULE', '00 01 * * *')
     placeholders.setdefault('CRONTAB', '[]')
     placeholders.setdefault('PGROOT', os.path.join(placeholders['PGHOME'], 'pgroot'))
+    placeholders.setdefault('WALE_TMPDIR', os.path.abspath(os.path.join(placeholders['PGROOT'], '../tmp')))
```
avaczi: How much space does WAL-E normally consume in the temp dir?  should we be concerned about small disk volumes on the root partition?

avaczi:
:+1:

postgres-appliance/configure_spilo.py
CyberDem0n: It uses TMPDIR for keeping produced files there before uploading them to S3.
Total size of these files during backup could be quite big. Every file up to 1GB (lzo algorithm is fast but not very efficient). Number of parallel streams depends on amount of CPU cores.
But the weird thing is - sometimes something goes wrong (probably wal-e crashes) and temp files stays there forever...

alexeyklyukin:
👍

PR #112, opened 2016-12-13
Introduce new postgresql parameters.

alexeyklyukin:
- Move all parameters from senza base to spilo. This leaves only
  one place to change their default values. We can still alter them
  from the template.
- Move filesystem-specific parameters out of the bootstrap section.
- Customize checkpoint and autovacuum parameters.


CyberDem0n:
:+1:

alexeyklyukin:
👍

PR #120, opened 2017-01-13
Some optimisations around restore_command = wal-e

CyberDem0n:
1) wal-e wal-fetch is incredibly slow
2) When we start postgres it will call the restore_command for every
xlog segment needed to reach consistent state even if there is a valid
file in the pg_xlog dicrectory.
3) Even if the wal-e is prefetching wal segments it it is rather
expencive to move them into pg_xlog by wal-e.

All these facts are making useless streaming of xlogs with pg_basebackup,
because afterwards it will call restore_command anyway :(

In order to optimize this process we will stream xlogs not into default
pg_xlog but into separate directory and create a wrapper restore_command
script, which will:
Check if there is xlog already available on disk either in directory
created by pg_basebackup or in the pg_xlog/.wal-e/prefetch directory.
If the file is there - just move it. If it is not - call wal-e with
usual options.

### DIFF
--- a/postgres-appliance/basebackup.sh
+++ b/postgres-appliance/basebackup.sh
@@ -34,6 +34,7 @@ while [[ $((ATTEMPT++)) -le $RETRIES ]]; do
     EXITCODE=$?
     if [[ $EXITCODE == 0 ]]; then
         XLOG_FAST=$(dirname $DATA_DIR)/xlog_fast
+        rm -fr $XLOG_FAST
         mv ${DATA_DIR}/pg_xlog $XLOG_FAST
         rm -fr $XLOG_FAST/archive_status
         break
--- a/postgres-appliance/basebackup.sh
+++ b/postgres-appliance/basebackup.sh
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+PARSED=$(getopt --options v --longoptions connstring:,retries:,datadir: --name "$0" -- "$@" 2> /dev/null)
+eval set -- "$PARSED"
+
 RETRIES=2
 
 while [[ $# -gt 0 ]]; do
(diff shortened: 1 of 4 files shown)


postgres-appliance/configure_spilo.py
```diff
@@ -142,7 +142,7 @@ def deep_update(a, b):
         autovacuum_analyze_scale_factor: 0.02
       {{#USE_WALE}}
       recovery_conf:
-        restore_command: envdir "{{WALE_ENV_DIR}}" wal-e --aws-instance-profile wal-fetch "%f" "%p"
+        restore_command: envdir "{{WALE_ENV_DIR}}" /wale_restore_command.sh
```
a1exsh: Where did `%f` and `%p` go?

postgres-appliance/basebackup.sh
```diff
@@ -0,0 +1,36 @@
...
+            ;;
+        --connstring )
+            CONNSTR=$2
+            shift
+            ;;
+        --retries )
+            RETRIES=$2
+            shift
+            ;;
+        * )
```
a1exsh: Should we rather error out here?

postgres-appliance/basebackup.sh
```diff
@@ -0,0 +1,36 @@
...
+
+[[ -z $DATA_DIR || -z $CONNSTR || ! $RETRIES =~ ^[1-9]$ ]] && exit 1
+
+DATA_DIR=$(realpath $DATA_DIR)
+XLOG_DIR=$(dirname $DATA_DIR)/xlog_fast
+
+ATTEMPT=0
+EXITCODE=1
+while [[ $((ATTEMPT++)) < $RETRIES || $EXITCODE == 0 ]]; do
+    pg_basebackup -P --pgdata="${DATA_DIR}" --xlog-method=stream --xlogdir="${XLOG_DIR}" --dbname="${CONNSTR}"
```
alexeyklyukin: -P attempts to shows progress information in this non-interactive script. It actually slows down the base backup, as it tries to figure out the amount of work to do before the backup starts. And it is imprecise when xlog segments are being streamed. So we should avoid it here.

postgres-appliance/basebackup.sh
CyberDem0n: Not really, there might be some other arguments passed by Patroni

alexeyklyukin:
👍

PR #124, opened 2017-01-25
Give RW permissions to WALE_TMPDIR for all

CyberDem0n:


postgres-appliance/configure_spilo.py
```diff
@@ -389,7 +389,7 @@ def write_wale_command_environment(placeholders, overwrite, provider):
     else:
         return
     if not os.path.exists(placeholders['WALE_TMPDIR']):
-        os.makedirs(placeholders['WALE_TMPDIR'])
+        os.makedirs(placeholders['WALE_TMPDIR'], mode=0777)
```
avaczi: Why is world readable/writable needed?

CyberDem0n: configure_spilo.py is executed by root and therefore directory is writable only by root, but we want to give access there to the postgres user as well.

Actually I've also forgot about "sticky bit".

avaczi:
:+1:

CyberDem0n:
:+1:

PR #129, opened 2017-02-07
Quote some of the parameters in postgres.yml

CyberDem0n:
And do small refactoring of the code which sets some other parameter values.


a1exsh:
:+1:

CyberDem0n:
:+1:

PR #132, opened 2017-02-16
Update README.rst

LappleApple:


README.md
```diff
@@ -1 +1,44 @@
-The documentation is on [spilo.readthedocs.org](http://spilo.readthedocs.org/).
+## Spilo: HA PostgreSQL Clusters with Docker
+
+Spilo is a Docker image that provides a resilient, High Available PostgreSQL cluster (HA-cluster) that you can configure and start within minutes. Spilo can run several clusters at one time, with one cluster acting as a master and the others as slaves. Its name derives from სპილო [spiːlɒ], the Georgian word for "elephant."
```
alexeyklyukin: Spilo provides PostgreSQL and Patroni bundled together. Multiple Spilos can create a HA PostgreSQL cluster. For this, all participating Spilos should be started with the same DCS (distributed configuration system) and cluster name parameters.

README.md
```diff
@@ -1 +1,44 @@
...
+
+Spilo works in combination with [Patroni](https://github.com/zalando/patroni), a template for PostgreSQL HA. Patroni works with [etcd](https://coreos.com/etcd/), [ZooKeeper](https://zookeeper.apache.org/) or [Consul](https://www.consul.io/), and manages the [PostgreSQL](https://www.postgresql.org) databases running in your Spilo cluster. It ensures that there is only one master within your Spilo. (It's worth noting here that Patroni includes a Helm Chart that enables you to deploy a five-node Patroni cluster using a Kubernetes PetSet, in a Google Compute Engine environment.)
+
+###How to Use This Docker Image
+
+First, there are some technical requirements:
+- A Distributed Configuration Store (DCS), to allow distributed nodes to agree. Most Patroni deployments use etcd, which implements the [Raft protocol](https://github.com/coreos/etcd/tree/master/raft). For more information about Raft, visit [this resource](http://thesecretlivesofdata.com/raft/) and check out [this interactive visualization](https://raft.github.io).
```
alexeyklyukin: I think we don't need to explain what RAFT here, this is the task of Etcd developers :-)

README.md
```diff
@@ -1 +1,44 @@
...
+
+>>>**not sure if this is correct?** This image includes 5432 (the Postgres port). The default Postgres user and database are created in the entrypoint with initdb.
```
alexeyklyukin: This sentence belongs to the administrative connections section. Instead of the first sentence, It should say that PostgreSQL is configured by default to listen to port 5432

README.md
```diff
@@ -1 +1,44 @@
...
+
+#### Administrative Connections//[could this be put as "Connecting to Postgres"?]
```
alexeyklyukin: Yes, both this and the Application Connections should be a subsection of "Connecting to Postgres".

README.md
```diff
@@ -1 +1,44 @@
...
+#### Application Connections
+
+Once you have created a database and roles for your application, you can connect to Spilo just like you want to connect to any other PostgreSQL cluster:
+
+    psql -h myfirstspilo.example.com -p 5432 -U wow_app -d wow
+    "postgresql://myfirstspilo.example.com:5432/wow?user=wow_app"
+
+#### Connecting to Other Containers
+
+#### Environment Variables
```
alexeyklyukin: We have those environment variables that needs to be documented, as without them none will be do anything interesting with WAL-E.

ETCD_HOST - the A record pointing to Etcd hosts. 
ETCD_DISCOVERY_DOMAIN - the SRC record pointing to Etcd hosts.
PGHOME - filesystem path where to put PostgreSQL home directory (/home/postgres by default)
APIPORT - TCP port to Patroni API connections (8008 by default)
BACKUP_SCHEDULE - cron schedule for doing backups via WAL-E (if WAL-E is enabled, '00 01 * * *' by default)
CRONTAB - anything that you want to run periodically as a cron job (empty by default)
PGROOT - a directory where we put the pgdata (by default /home/postgres/pgroot). One may adjust it to point to the mount point of the persistent volume, such as EBS.
WALE_TMPDIR - directory to store WAL-E temporary files.  $PGROOT/../tmp by default, make sure it has a few GBs of free space.
PGDATA - location of PostgreSQL data directory, by default $PGROOT/pgdata.
PGPASSWORD_STANDBY - a password for the standby user, 'standby' by default.
PGPASSWORD_SUPERUSER - a password for the superuser, 'zalando' by default
PGPORT - port PostgreSQL listens to for client connections, 5432 by default
SCOPE - cluster name, multiple Spilos belonging to the same cluster must have identical scope.
SSL_CERTIFICATE_FILE - path to the SSL certificate file inside the container (by default PGHOME/server.crt), Spilo will generate one if not present.
SSL_PRIVATE_KEY_FILE - path to the SSL private key within the container (by default PGHOME/server.key), Spilo will generate one if not present
WALE_BACKUP_THRESHOLD_MEGABYTES - maximum size of the WAL segments accumulated after the base backup to consider WAL-E restore instead of pg_basebackup.
WALE_BACKUP_THRESHOLD_PERCENTAGE - maximum ratio (in percents) of the accumulated WAL files to the base backup to consider WAL-E restore instead of pg_basebackup.
WALE_ENV_DIR - directory where to store WAL-E environment variables
WAL_S3_BUCKET - path to the S3 bucket used for WAL-E base backups (i.e. s3://foobar). Spilo will add /spilo/$scope/wal to that path.
WAL_GCS_BUCKET - ditto for the Google Cloud Storage (WAL-E supports both S3 and GCS).
USE_WALE - whether to consider base backup/restore with WAL-E.
GOOGLE_APPLICATION_CREDENTIALS - credentials for WAL-E when running in Google Cloud.
CALLBACK_SCRIPT - the callback script to run on various cluster actions (on start, on stop, on restart, on role change). The script will receive the cluster name, connection string and the current action. See Patroni documentation for details

README.md
alexeyklyukin: One can replace DCS to Etcd, since Spilo supports only Etcd at the moment.

README.md
```diff
@@ -1 +1,44 @@
...
+#### Supported Tags and Respective Dockerfile Links
+#### How to Extend This Image
```
alexeyklyukin: Redundant with "Contributions"

README.md
```diff
@@ -1 +1,44 @@
...
+
+### Connect to a Spilo Appliance
```
alexeyklyukin: Redundant with "PostgreSQL connections".

README.md
alexeyklyukin: Also, add that Spilo requires a load balancer (Haproxy, ELB, Google load balancer) to point to the master node. It provides 2 ways to achieve this: either using the API URLs, or via the callback scripts. See Patroni project for details.

README.md
```diff
@@ -1 +1,44 @@
...
+
+Spilo is a Docker image that provides a resilient, High Available PostgreSQL cluster (HA-cluster) that you can configure and start within minutes. Spilo can run several clusters at one time, with one cluster acting as a master and the others as slaves. Its name derives from სპილო [spiːlɒ], the Georgian word for "elephant." 
+
+Spilo is currently evolving: Its creators are working on a Postgres operator that would make it simpler to deploy scalable Postgres clusters in a Kubernetes environment, and do maintenance tasks. Spilo would serve as an essential building block for this.
+
+Spilo works in combination with [Patroni](https://github.com/zalando/patroni), a template for PostgreSQL HA. Patroni works with [etcd](https://coreos.com/etcd/), [ZooKeeper](https://zookeeper.apache.org/) or [Consul](https://www.consul.io/), and manages the [PostgreSQL](https://www.postgresql.org) databases running in your Spilo cluster. It ensures that there is only one master within your Spilo. (It's worth noting here that Patroni includes a Helm Chart that enables you to deploy a five-node Patroni cluster using a Kubernetes PetSet, in a Google Compute Engine environment.)
```
LappleApple: From the Patroni README: " To this end, we've created a Helm Chart that enables you to deploy a five-node Patroni cluster using a Kubernetes PetSet." 

The word "includes" means that "there is a Helm chart for Patroni that you can use."

alexeyklyukin: Yeah, the Helm chart is actually named wrongly, since it's deploying Spilo docker images, not just "bare" Patroni.  At some point we tried to avoid using Spilo for anything that is not STUPS, that resulted in the confusion with the names. I'd put the Helm chart mention into the separate chapter (Real-word usage or something like this), where we can mention Helm, Stups (or maybe not) and say some words about upcoming operator development. 

I'd also phrase the Patroni - Helm chart relationship differently, i.e. there is a Helm chart that relies on Spilo and Patroni to provision a five-node PostgreSQL HA cluster.

LappleApple:
@alexeyklyukin I made some changes based on your comments, and also converted from .md to .rst. Take a look :)

README.rst
```diff
@@ -0,0 +1,47 @@
+==================================
+Spilo: HA PostgreSQL Clusters with Docker
+==================================
+
+Spilo is a Docker image that provides PostgreSQL and `Patroni <https://github.com/zalando/patroni>`__ bundled together. Patroni is a template for PostgreSQL HA. Multiple Spilos can create a resilient High Available PostgreSQL cluster. For this, you'll need to start all participating Spilos with `etcd <https://github.com/coreos/etcd>`__ and cluster name parameters. Spilo's name derives from სპილო [spiːlɒ], the Georgian word for "elephant."
```
alexeyklyukin: not just with etcd and cluster name parameters, but with identical ones, i.e. all Spilos participating in a single HA cluster should point to the same Etcd and have the same cluster name (which is called scope in Spilo).

README.rst
```diff
@@ -0,0 +1,47 @@
...
+
+Real-World Usage and Plans
+--------------------------
+
+Spilo is currently evolving: Its creators are working on a Postgres operator that would make it simpler to deploy scalable Postgres clusters in a Kubernetes environment, and also do maintenance tasks. Spilo would serve as an essential building block for this. There is already a `Helm chart <https://github.com/kubernetes/charts/tree/master/incubator/patroni>`__ that relies on Spilo and Patroni to provision a five-node PostgreSQL HA cluster in a Kubernetes+Google Compute Engine environment. (The Helm chart deploys Spilo Docker images, not just "bare" Patroni.)
+
+How to Use This Docker Image
+============================
+
+Spilo requires a load balancer (HAProxy, ELB, Google load balancer) to point to the master node. It provides two ways to achieve this: either by using the API URLs, or via the callback scripts. See Patroni for more details.
```
alexeyklyukin: Spilo doesn't require a load balancer: the client connecting to Spilo needs it. The problem, from the client perspective, is that there are N Spilos in the HA Postgres cluster, only one of them is "active" (i.e. master), e.g. accepting writes. How do we connect the active one? For that, we need a service in front of the HA cluster, providing a single connection point to the clients and then forwarding that connection to the master node. It could be the ELB, Google load balancer, HA Proxy, Kubernetes Endpoints or anything else. Spilo, via the underlying Patroni, has 2 mechanisms to inform that service about the location of the master node: either via the callbacks (http://patroni.readthedocs.io/en/latest/SETTINGS.html?highlight=callback#postgresql) or via the REST API endpoints, listening on the *APIPORT* on the address of the host (or pod) spilo runs at. Those endpoints return 500 for the replica and 200 for the master Spilo.

README.rst
```diff
@@ -0,0 +1,47 @@
...
+
+Connecting to PostgreSQL
+------------------------
+**Administrative Connections**
+
+PostgreSQL is configured by default to listen to port 5432. The default Postgres user and database are created in the entrypoint with initdb.
```
alexeyklyukin: The second sentence is true, but doesn't add any value to the documentation. When the cluster is bootstrapped, Spilo master initializes PostgreSQL with `initdb` and creates the superuser and replication user (`postgres` and `standby` by default).

README.rst
```diff
@@ -0,0 +1,47 @@
...
+You'll need to setup Spilo to create a database and roles for your application(s). For example:
+
+    psql -h myfirstspilo.example.com -p 5432 -U admin -d postgres
+
+**Application Connections**
+
+Once you have created a database and roles for your application, you can connect to Spilo just like you want to connect to any other PostgreSQL cluster:
+
+    psql -h myfirstspilo.example.com -p 5432 -U wow_app -d wow
+    "postgresql://myfirstspilo.example.com:5432/wow?user=wow_app"
```
alexeyklyukin: it should be psql -d  "postgresql://myfirstspilo.example.com:5432/wow?user=wow_app"

README.rst
```diff
@@ -0,0 +1,47 @@
...
+
+Environment Variables
```
alexeyklyukin: I think we need to rename it to "Configuration" and add a sentence, saying that Spilo is configured via environment variables. The values for those variables are either supplied manually via the environment, when Spilo is launched as a set of Docker containers, or added in the configuration file or manifest, when Spilo is used in the docker orchestration environment, such as Kubernetes or Docker Compose.

README.rst
alexeyklyukin: A space is missing, since those should be separate lines.

alexeyklyukin: psql -d "postgresql://myfirstspilo.example.com:5432/wow?user=wow_app"

README.rst
LappleApple: My text here is based on your earlier comment, which didn't mention identical name parameters. My suggestion is that you rewrite this paragraph to state your complete meaning, and I'll edit it.

README.rst
LappleApple: This was your comment from earlier: "Also, add that Spilo requires a load balancer (Haproxy, ELB, Google load balancer) to point to the master node. It provides 2 ways to achieve this: either using the API URLs, or via the callback scripts. See Patroni project for details."

README.rst
LappleApple: I've seen it used elsewhere; sometimes overcommunication is not a bad idea.

README.rst
alexeyklyukin: My comment mentioned *the same*  DCS (distributed configuration system) and cluster name parameters. I'd change your paragraph like this:

Spilo is a Docker image that provides PostgreSQL and `Patroni <https://github.com/zalando/patroni>`__ bundled together. Patroni is a template for PostgreSQL HA. Multiple Spilos can create a resilient High Available PostgreSQL cluster. For this, you'll need to start all participating Spilos with identical Etcd addresses and cluster names. Spilo's name derives from სპილო [spiːlɒ], the Georgian word for "elephant."

README.rst
LappleApple: fixed

LappleApple: fixed

LappleApple: fixed

README.rst
LappleApple: done

README.rst
LappleApple: LGTM; I'll make the change inside the file

README.rst
alexeyklyukin: Yeah, it was a poor choice of words from my side. Spilo relies on a correctly configured load balancer (...) to direct client connections to the master. There are 2 ways to achieve this: if the load balancer relies on the status code to distinguish between the healthy and failed nodes (like ELB), then one needs to configure it to poll the API URL. Otherwise, one can use callback scripts to change the load balancer configuration dynamically.

LappleApple: Hey, that's fine. This version of the text is a lot clearer, good work

README.rst
alexeyklyukin: The reason it doesn't add any value is because it is quite vague: it refers to the Docker image entry point, which is unnecessary implementation detail, and mentions that default Postgres users and database is created with initdb, and this is *always* the case, no matter with Spilo or don't. Therefore, I think we should change it to 

"Spilo master initializes PostgreSQL and creates the superuser and replication user (`postgres` and `standby` by default)."

LappleApple: LGTM; fixed

README.rst
```diff
@@ -0,0 +1,51 @@
...
+
+    psql -h myfirstspilo.example.com -p 5432 -U wow_app -d wow
+    psql -d "postgresql://myfirstspilo.example.com:5432/wow?user=wow_app"
+
+Configuration
+-------------
+
+Spilo is configured via environment variables, the values of which are either supplied manually via the environment (whenever Spilo is launched as a set of Docker containers) or added in the configuration file or manifest (whenever Spilo is used in the Docker orchestration environment, such as Kubernetes or Docker Compose).
+
+Please go `here <hURL TO COME>`__ to see our list of environment variables.
```
alexeyklyukin: Here comes the URL: https://github.com/zalando/spilo/blob/master/ENVIRONMENT.rst

alexeyklyukin:
👍

LappleApple:
👍

alexeyklyukin:
#132

alexeyklyukin:
Excellent  work, thanks a lot! @LappleApple.

PR #134, opened 2017-02-16
Enable ssl compression by recompiling libssl

CyberDem0n:
And make compression level = Z_BEST_SPEED (better level doesn't bring much benefit but CPU greedy)


alexeyklyukin:
👍

a1exsh:
:+1:

PR #135, opened 2017-02-19
Install pg_repack and some refactoring

CyberDem0n:
* move rebuild/install of libssl to upper layer
* move build/install of pam-ouath2 to upper layer
* install postgresql and 3rd party extensions from one layer, starting from postgres 9.3 and up to 9.6.  It's necessary, because otherwise it's not possible to build some additional extensions. Extensions are relying on libpq-dev from the actual version, but usually only most recent version is installed and it's not possible to downgrade it.
* install pg_rewind on 9.3 and 9.4
* install pg_repack from github + PR#107, otherwise it's failing on tables with upserts. We should switch back to pgxn when they will release a new version with the fix.
* increase amount of retries in boto when fetching instance metadata

### DIFF
--- a/postgres-appliance/Dockerfile
+++ b/postgres-appliance/Dockerfile
@@ -1,19 +1,45 @@
 FROM ubuntu:14.04
 MAINTAINER Oleksii Kliukin <oleksii.kliukin@zalando.de>
 
-# Install curl, jq, vim, gdb, strace
 RUN export DEBIAN_FRONTEND=noninteractive \
+    && export BUILD_PACKAGES="build-essential dpkg-dev devscripts debhelper fakeroot m4 bc bsdmainutils libcurl4-openssl-dev libpam0g-dev" \
     && apt-get update \
     && echo 'APT::Install-Recommends "0";' > /etc/apt/apt.conf.d/01norecommend \
     && echo 'APT::Install-Suggests "0";' >> /etc/apt/apt.conf.d/01norecommend \
 
     && apt-get upgrade -y \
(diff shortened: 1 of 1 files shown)


postgres-appliance/Dockerfile
```diff
@@ -64,31 +122,13 @@ RUN export DEBIAN_FRONTEND=noninteractive \
...
-    done \
-    && export PATH=$OLD_PATH \
-
     && pip3 install pip --upgrade \
     && pip3 install --upgrade packaging appdirs requests pystache patroni==$PATRONIVERSION \
             gcloud boto wal-e==$WALE_VERSION \
-    && for extension in ${PGXN_EXTENSIONS}; do pgxn install $extension; done \
+
+    # https://github.com/wal-e/wal-e/issues/318
+    && sed -i 's/^\(    for i in range(0,\) num_retries):.*/\1 100):/g' /usr/local/lib/python3.4/dist-packages/boto/utils.py \
```
a1exsh: It seems that there is a supported way to get what you want here, see `metadata_service_num_attempts`: http://docs.aws.amazon.com/cli/latest/reference/configure/#configuration-variables

a1exsh:
I believe this is better split into individual commits or even separate pull requests.  Too many changes at a time, IMO.

postgres-appliance/Dockerfile
alexeyklyukin: The page you've linked to says it's only used for AWS CLI. The boto (2) page explicitly lists configuration files boto searches: http://docs.pythonboto.org/en/latest/boto_config_tut.html

a1exsh:
:+1:

CyberDem0n:
:+1:

PR #140, opened 2017-03-15
Update manifest for Spilo.

alexeyklyukin:
Includes S3 WAL shipping and base backups.

### DIFF
--- a/kubernetes/spilo_kubernetes_aws.yaml
+++ b/kubernetes/spilo_kubernetes_aws.yaml
@@ -1,0 +1,155 @@
+apiVersion: apps/v1beta1
+kind: StatefulSet
+metadata:
+  name: &cluster_name zalandodemo01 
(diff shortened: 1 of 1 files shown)


kubernetes/spilo.yaml
```diff
@@ -0,0 +1,146 @@
...
+                            }
+                          ]
+                         },
+                         "topologyKey": "kubernetes.io/hostname" 
+                    }
+                  ]
+                 }
+            }
+    spec:
+      serviceAccountName: operator
```
alexeyklyukin: Hm, probably not. I don't want to remove it, but I will add a comemnt, explaining, what does it do and why do we need it.

alexeyklyukin:
👍

CyberDem0n:
:+1:

alexeyklyukin:
👍

PR #143, opened 2017-04-04
Make number of backups to retain configurable using environment

a1exsh:
The new env var name follows doc of wal-e "delete retain" subcommand.

### DIFF
--- a/postgres-appliance/postgres_backup.sh
+++ b/postgres-appliance/postgres_backup.sh
@@ -16,12 +16,17 @@ shift
 PGDATA=$1
 shift
 
+NUM_TO_RETAIN=$1
+shift
+
 IN_RECOVERY=$(psql -tXqAc "select pg_is_in_recovery()")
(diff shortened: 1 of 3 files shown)


postgres-appliance/postgres_backup.sh
```diff
@@ -16,12 +16,17 @@ shift
...
 
+NUM_TO_RETAIN=$1
+shift
+
 IN_RECOVERY=$(psql -tXqAc "select pg_is_in_recovery()")
 [[ $IN_RECOVERY != "f" ]] && log "Cluster is in recovery, not running backup" && exit 0
 
-# leave only 2 base backups before creating a new one
+# leave at least 2 base backups before creating a new one
+[[ "$NUM_TO_RETAIN" -lt 2 ]] && NUM_TO_RETAIN=2
```
a1exsh: Hm, or maybe we should just rely on default being set to 2 by initial configuration script...

alexeyklyukin: Perhaps we should safeguard only against this value being less than one.

alexeyklyukin:
👍

PR #145, opened 2017-04-11
Grant access to postgres_log partitions to admin

CyberDem0n:
It's not nice to query parent table because it will have to parse 7 or 8 csv files all the time.


alexeyklyukin:
👍

avaczi:
:+1:

PR #147, opened 2017-04-12
v prefix is not needed when we are refering commit id

CyberDem0n:
it is only necessary when we reference a specific tag.


alexeyklyukin:
👍

CyberDem0n:
:+1:

PR #150, opened 2017-04-19
Make sure 'locales' package is installed

joar:
The most recent ubuntu:16.04 image
(ubuntu@sha256:c2bbf50d276508d73dd865cda7b4ee9b5243f2648647d21e3a471dd3cc4209a0)
does not have the "locales" package installed, which results in the following
error message:

     character map file `UTF-8' not found: No such file or directory
     cannot read character map directory `/usr/share/i18n/charmaps': No such file or directory


alexeyklyukin:
👍

hjacobs:
:+1:

PR #154, opened 2017-04-24
Run pg_receivexlog until postgres will start streaming from master

CyberDem0n:
On heavily loaded clusters wal-e can't keep up with the rate of generated wal segments.

### DIFF
--- a/postgres-appliance/configure_spilo.py
+++ b/postgres-appliance/configure_spilo.py
@@ -188,15 +188,10 @@ postgresql:
     on_restart: {{CALLBACK_SCRIPT}}
     on_role_change: {{CALLBACK_SCRIPT}}
  {{/CALLBACK_SCRIPT}}
+{{#USE_WALE}}
   create_replica_method:
-    {{#USE_WALE}}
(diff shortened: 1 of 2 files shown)


postgres-appliance/configure_spilo.py
```diff
@@ -188,15 +188,10 @@ def deep_update(a, b):
...
     on_role_change: {{CALLBACK_SCRIPT}}
  {{/CALLBACK_SCRIPT}}
+{{#USE_WALE}}
   create_replica_method:
-    {{#USE_WALE}}
     - wal_e
     - basebackup_fast_xlog
-    {{/USE_WALE}}
-    {{^USE_WALE}}
-    - basebackup
```
a1exsh: Why remove this option completely?

CyberDem0n: Patroni will default to basebackup/pg_basebackup if there is no `create_replica_method:` defined in the config.

a1exsh:
:+1:

alexeyklyukin:
👍

PR #155, opened 2017-04-24
BUGFIX: get available memory from memory.limit_in_bytes

CyberDem0n:
/sys/fs/cgroup/memory/memory.max_usage_in_bytes records the maximum of memory usage


a1exsh:
:+1:

alexeyklyukin:
👍

PR #160, opened 2017-04-28
fix: Can't mix manual and automatic field numbering

joar:
Previously raised 
```pytb
  File "/configure_spilo.py", line 527, in <module>
    main()
  File "/configure_spilo.py", line 487, in main
    raise ValueError('{0} should contain a dict, yet it is a {}'.format(config_var_name, type(user_config)))
ValueError: cannot switch from manual field specification to automatic field numbering
```


CyberDem0n:
:+1:

alexeyklyukin:
👍

alexeyklyukin:
Merged, thank you!

PR #162, opened 2017-05-08
Reduce verify to 2 and check against /etc/ssl/certs

CyberDem0n:
Checking against locally installed certificate (CAfile) makes it impossible to switch server certificate without touching stunnel configuration.


alexeyklyukin:
👍

CyberDem0n:
:+1:

PR #165, opened 2017-05-15
Remove xlog segments created by pg_receivexlog.

alexeyklyukin:
Avoid leaving them around once the replica has started streaming.

### DIFF
--- a/postgres-appliance/basebackup.sh
+++ b/postgres-appliance/basebackup.sh
@@ -37,7 +37,7 @@ function receivexlog() {
         kill -0 $receivexlog_pid && sleep 1 || exit
     done
 
-    kill $receivexlog_pid && rm -fr ${XLOG_FAST}/*
+    kill $receivexlog_pid && sleep 1 && rm -f ${XLOG_FAST}/*
 }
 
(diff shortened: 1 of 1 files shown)


alexeyklyukin:
👍

postgres-appliance/basebackup.sh
```diff
@@ -37,7 +37,7 @@ function receivexlog() {
         kill -0 $receivexlog_pid && sleep 1 || exit
     done
 
-    kill $receivexlog_pid
+    kill $receivexlog_pid && rm -fr ${XLOG_FAST}/*
```
CyberDem0n: I would actually give a time to pg_receivexlog to exit, lets say 1 second.
And we don't need to remove recursively, `rm -f ${XLOG_FAST}/*` should be enough.

alexeyklyukin: Updated, thank you.

alexeyklyukin:
👍

CyberDem0n:
:+1:

zafinxueqian:
In Dockerfile line161, where is scm-source.json file

PR #171, opened 2017-06-20
Avoid calling aws callback on restart.

alexeyklyukin:
Restart doesn't change AWS labels of cluster members.

### DIFF
--- a/postgres-appliance/configure_spilo.py
+++ b/postgres-appliance/configure_spilo.py
@@ -186,9 +186,6 @@ postgresql:
   callbacks:
     on_start: {{CALLBACK_SCRIPT}}
     on_stop: {{CALLBACK_SCRIPT}}
-    {{#CALLBACK_RESTART_SCRIPT}}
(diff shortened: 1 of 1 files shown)


postgres-appliance/configure_spilo.py
```diff
@@ -289,7 +292,7 @@ def get_placeholders(provider):
 
     # Kubernetes requires a callback to change the labels in order to point to the new master
     if USE_KUBERNETES:
-        placeholders['CALLBACK_SCRIPT'] = '/callback_role.py'
+        placeholders['CALLBACK_SCRIPT'] = placeholders['CALLBACK_RESTART_SCRIPT'] = '/callback_role.py'
```
CyberDem0n: Why do we need to call on_restart on Kubernetes?

alexeyklyukin: Actually, I think we don't. Will change it as well.

CyberDem0n:
:+1:

alexeyklyukin:
👍

PR #175, opened 2017-07-18
New callback for associating elastic ip with master instance

CyberDem0n:
Also it does remaining tasks of patroni_aws (tag instances and volumes)

### DIFF
--- a/postgres-appliance/callback_aws.py
+++ b/postgres-appliance/callback_aws.py
@@ -1,0 +1,89 @@
+#!/usr/bin/env python
+
+import boto3
+import logging
+import requests
+import sys
+import time
+
+from botocore.exceptions import ClientError
+from requests.exceptions import RequestException
+
+
+logger = logging.getLogger(__name__)
+
+
+def retry(func):
+    def wrapped(*args, **kwargs):
+        count = 0
+        while True:
+            try:
+                return func(*args, **kwargs)
+            except ClientError as e:
+                if not (e.response['Error']['Code'] == 'Throttling' or 'RequestLimitExceeded' in str(e)):
+                    raise
+                logger.info('Throttling AWS API requests...')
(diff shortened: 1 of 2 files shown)


postgres-appliance/callback_aws.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+def retry(func):
+    def wrapped(*args, **kwargs):
+        count = 0
+        while True:
+            try:
+                return func(*args, **kwargs)
+            except ClientError as e:
+                if e.response['Error']['Code'] != 'Throttling' or 'RequestLimitExceeded' not in str(e):
```
alexeyklyukin: this is quite confusing and   is better to be spelled as ``not ( e.response['Error']['Code'] == 'Throttling' or 'RequestlimitExceeded' in str(e))``. Makes it clearer that it should be and in the original condition.

postgres-appliance/callback_aws.py
```diff
@@ -0,0 +1,89 @@
...
+
+    action, role, cluster = sys.argv[2:5]
+
+    metadata = get_instance_metadata()
+
+    instance_id = metadata['instanceId']
+
+    ec2 = boto3.client('ec2', region_name=metadata['region'])
+
+    if role == 'master' and action in ('on_start', 'on_role_change'):
```
alexeyklyukin: Should we also disassociate it when the instance is demoted? We might be running without the master for some time, i.e. if no suitable candidates are present.

CyberDem0n: I don't think that we should care about it.
Elastic IP could be associated only with one instance at a time and it can be associated without disassociation.

alexeyklyukin:
👍

CyberDem0n:
:+1:

PR #179, opened 2017-08-01
Don't override Name tags of EBS volumes

a1exsh:
Sometime we rely on Name tag of the data EBS volume, which is different from
the usual 'spilo_{cluster_name}' format, for example, when attaching multiple
volumes and making a RAID on top of them.

This will ensure that any pre-defined Name tags are kept intact.  The instance
and role tags are updated in any case.

### DIFF
--- a/postgres-appliance/callback_aws.py
+++ b/postgres-appliance/callback_aws.py
@@ -44,6 +44,11 @@ def list_volumes(ec2, instance_id):
     return ec2.get_all_volumes(filters={'attachment.instance-id': instance_id})
 
 
+@retry
+def get_instance(ec2, instance_id):
+    return ec2.get_only_instances([instance_id])[0]
+
+
 def main():
     logging.basicConfig(format='%(asctime)s %(levelname)s: %(message)s', level=logging.INFO)
     if len(sys.argv) != 5 or sys.argv[2] not in ('on_start', 'on_stop', 'on_role_change'):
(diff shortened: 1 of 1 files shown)


alexeyklyukin:
Does it make sense to pick another tag for the spilo cluster name?

postgres-appliance/callback_aws.py
```diff
@@ -67,11 +61,15 @@ def main():
...
+    tag_resource(ec2, instance_id, tags)
+
+    tags.update({'Instance': instance_id})
+    name_tag = 'spilo_' + cluster
 
-    tags.update({'Instance': instance_id, 'Name': 'spilo_' + cluster})
     volumes = list_volumes(ec2, instance_id)
-    tag_volumes(ec2, volumes, tags)
+    for v in volumes:
+        tags_to_update = tags if 'Name' in v.tags else dict(tags, Name=name_tag)
```
CyberDem0n: Should we differentiate between root volume and data volumes?

a1exsh: Well, we didn't do that before.  If we start now, that will introduce inconsistencies.

a1exsh:
@alexeyklyukin currently the Name tag is only used for sort/filtering in the AWS console.  If we would use a different tag name, the console will need to be customized by every user to display it.

alexeyklyukin:
@a1exsh but if we don't - wouldn't it lead to inconsistent results when filtering/sorting on that tag?

avaczi:
From a billing perspective, the spilo cluster name (and a tag for the team, but that might be too complicated to solve on this level) would be a very good idea @alexeyklyukin

CyberDem0n:
:+1:

a1exsh:
:+1:

PR #180, opened 2017-08-02
Don't try to start pgbouncer in the background under supervisord

a1exsh:
On the first attempt to start it will succeed and exit, but continue running
in the background, so when supervisord will try to start it again it will be
constantly failing because the unix socket is already used.


ikitiki:
👍

alexeyklyukin:
👍

PR #181, opened 2017-08-04
Custom bootstrap scripts for Patroni

alexeyklyukin:
WIP: Scripts for bootstrap with base backup and wal-e.


CyberDem0n:
Looks like you moved the wrong file, configure_spilo.py instead of clone_with_s3.py

alexeyklyukin:
The joy of last-minute changes. Moved back, thanks.

postgres-appliance/bootstrap/clone_with_basebackup.py
```diff
@@ -0,0 +1,76 @@
...
+import subprocess
+
+logging.basicConfig(format='%(asctime)s %(levelname)s: %(message)s', level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+def read_configuration():
+    parser = argparse.ArgumentParser(description="Script to clone from another cluster using pg_basebackup")
+    parser.add_argument('--scope', required=True, help='target cluster name')
+    parser.add_argument('--datadir', required=True, help='target cluster postgres data directory')
+    parser.add_argument('--bindir', required=True, help='the directory with pg_basebackup')
```
alexeyklyukin: Originally, those scripts were not supposed to be tied to Spilo, therefore, no guarantees PATH will be set (as it is set by the Spilo Docker image). While they are now part of Spilo, I'd prefer to make as few assumptions as possible on the environment they run at (they still assume wal-e is accessible in PATH, but that is a reasonable assumption for now).

postgres-appliance/bootstrap/clone_with_basebackup.py
```diff
@@ -0,0 +1,76 @@
...
+        if c.isspace():
+            quote = True
+        elif c in ('\'', '\\'):
+            temp.append('\\')
+        temp.append(c)
+    result = ''.join(temp)
+    return result if not quote else '\'{0}\''.format(result)
+
+def prepare_connection(options):
+    pgpass = parse_pgpass(options.pgpassfile)
```
alexeyklyukin: I use pgpass to pass the password for the connection, and since pgpass requires exact match of other fields as well, and those are exact fields required to connect to the cluster, it makes sense to read all of them from pgpass.

CyberDem0n: I think it's better to avoid unnecessary parsing, it's not easy to get it right. For example you code will break it hostname is in IPv6 format...

alexeyklyukin: Well, that should not be that hard, one has to take care of '\' and ':' characters, and they should have taken care of anyway when .pgpass is written (on the Spilo side).

postgres-appliance/bootstrap/clone_with_s3.py
```diff
@@ -0,0 +1,95 @@
...
+    backup_name = 'LATEST'
+    if options.recovery_target_time:
+        backup_list_cmd = build_wale_command(options.wale_envdir, 'backup-list')
+        backup_list = subprocess.check_output(backup_list_cmd)
+        backup_name = choose_backup(backup_list, options.recovery_target_time)
+    backup_fetch_cmd = build_wale_command(options.wale_envdir, 'backup-fetch', datadir=options.datadir, backup=backup_name)
+    logger.info("cloning cluster {0} using {1}".format(options.name, ' '.join(backup_fetch_cmd)))
+    if not options.dry_run:
+        ret = subprocess.call(backup_fetch_cmd)
+        if ret:
```
CyberDem0n: if ret != 0:

postgres-appliance/configure_spilo.py
```diff
@@ -140,6 +141,30 @@ def deep_update(a, b):
...
+        pause_at_recovery_target: false
+        {{/USE_PAUSE_AT_RECOVERY_TARGET}}
+        {{^USE_PAUSE_AT_RECOVERY_TARGET}}
+        recovery_target_action: promote
+        {{/USE_PAUSE_AT_RECOVERY_TARGET}}
+        {{#CLONE_TARGET_TIME}}
+        recovery_target_time: "{{CLONE_TARGET_TIME}}"
+        {{/CLONE_TARGET_TIME}}
+  {{/CLONE_WITH_WALE}}
+  {{^CLONE_WITH_WALE}}
```
CyberDem0n: `CLONE_WITH_BASEBACKUP` could only bet set if `CLONE_WITH_WALE` is not set.

postgres-appliance/configure_spilo.py
```diff
@@ -347,37 +412,78 @@ def get_dcs_config(config, placeholders):
...
 
+def write_wale_command_environment(placeholders, provider, overwrite):
+    if not placeholders['USE_WALE']:
+        return
+
+    write_wale_environment(placeholders, provider, '', overwrite)
+
+
+def write_wale_clone_environment(placeholders, provider, overwrite):
+    if not placeholders['CLONE_WITH_WALE']:
```
CyberDem0n: I think it make sense to move this check to the line https://github.com/zalando/spilo/pull/181/files#diff-9ec119124414a288bc2f1edd468dcce5R467 and get rid from `write_wale_clone_environment` function.

postgres-appliance/configure_spilo.py
```diff
@@ -347,37 +412,78 @@ def get_dcs_config(config, placeholders):
...
     if not os.path.exists(placeholders['WALE_TMPDIR']):
         os.makedirs(placeholders['WALE_TMPDIR'])
         os.chmod(placeholders['WALE_TMPDIR'], 0o1777)
 
-    write_file(placeholders['WALE_TMPDIR'], os.path.join(placeholders['WALE_ENV_DIR'], 'TMPDIR'), True)
+    write_file(placeholders['WALE_TMPDIR'], os.path.join(wale['WALE_ENV_DIR'], 'TMPDIR'), True)
+
 
+def write_wale_command_environment(placeholders, provider, overwrite):
+    if not placeholders['USE_WALE']:
```
CyberDem0n: I think it make sense to move this check to the line https://github.com/zalando/spilo/pull/181/files#diff-9ec119124414a288bc2f1edd468dcce5R647 and get rid from `write_wale_command_environment` function.

postgres-appliance/configure_spilo.py
```diff
@@ -347,37 +412,78 @@ def get_dcs_config(config, placeholders):
...
+    write_wale_environment(placeholders, provider, 'CLONE_', overwrite)
+
+
+def write_bootstrap_configuration(placeholders, provider, overwrite):
+    write_wale_clone_environment(placeholders, provider, overwrite)
+    write_clone_pgpass(placeholders, overwrite)
+
+
+def write_clone_pgpass(placeholders, overwrite):
+    if not placeholders['CLONE_WITH_BASEBACKUP']:
```
CyberDem0n: Move check to the line https://github.com/zalando/spilo/pull/181/files#diff-9ec119124414a288bc2f1edd468dcce5R468
to avoid calling function when it is not necessary.

postgres-appliance/bootstrap/clone_with_basebackup.py
```diff
@@ -0,0 +1,119 @@
...
+    connection = []
+    for attname in ('host', 'port', 'user', 'dbname'):
+        attvalue = getattr(pgpass, attname)
+        if attvalue and attvalue != '*':
+            connection.append('{0}={1}'.format(attname, escape_value(attvalue)))
+
+    return ' '.join(connection), {'PGPASSFILE': options.pgpassfile}
+
+def run_basebackup(options):
+    pg_basebackup = os.path.join(options.bindir, 'pg_basebackup')
```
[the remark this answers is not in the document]
alexeyklyukin: The ability to specify 'bindir' from the command-line simplifies testing. When it is omitted - the os.path.join will just produce 'pg_basebackup'.

postgres-appliance/bootstrap/clone_with_basebackup.py
```diff
@@ -0,0 +1,119 @@
...
+
+def prepare_connection(options):
+    pgpass = parse_pgpass_file(options.pgpassfile)
+    connection = []
+    for attname in ('host', 'port', 'user', 'dbname'):
+        attvalue = getattr(pgpass, attname)
+        if attvalue and attvalue != '*':
+            connection.append('{0}={1}'.format(attname, escape_value(attvalue)))
+
+    return ' '.join(connection), {'PGPASSFILE': options.pgpassfile}
```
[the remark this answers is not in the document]
alexeyklyukin: In our case we cannot be explicit, since passwords should not be passed in the open, and since .pgpass is a convenient way to handle passwords and it also contains the rest of the connection arguments we need to make a connection string  I see no reasons not to use it.

postgres-appliance/configure_spilo.py
alexeyklyukin: Unfortunately I couldn't find any way to express this in the template, and as the logic that guarantees that is in the Python code (that could be changed independently of the template) I thought of adding another check in the template itself. I see now it makes the template unreadable, so I'll get rid of it.

postgres-appliance/bootstrap/clone_with_basebackup.py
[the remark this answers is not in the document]
CyberDem0n: I am not asking to pass the password as a part of connection string.
We just need to pass two arguments:
1. connection string (without password)
2. location of .pgpass

It will simplify a lot of things and will allow to get rid from ~75 lines of code parsing .pgpass...

postgres-appliance/bootstrap/clone_with_basebackup.py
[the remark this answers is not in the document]
CyberDem0n: If there would be no bindir, it would be not necessary to use `os.path.join`.

And about tests... There is always a way to execute them:
`$ PATH=/path/to/postgres/bin:$PATH python clone_with_basebackup.py`

postgres-appliance/bootstrap/clone_with_s3.py
```diff
@@ -0,0 +1,95 @@
...
+    return cmd
+
+def choose_backup(output, t):
+    """ pick up the latest backup file starting before time t"""
+    reader = csv.DictReader(output.decode('utf-8').splitlines(), dialect='excel-tab')
+    backup_list = list(reader)
+    if len(backup_list) <= 0:
+        raise Exception("wal-e could not found any backups")
+    match = None
+    for i, backup in enumerate(backup_list):
```
CyberDem0n: `i` is not used here

```diff
-    for i, backup in enumerate(backup_list):
+    for backup in backup_list:
```

alexeyklyukin:
👍

CyberDem0n:
:+1:

PR #182, opened 2017-08-08
Patroni 1.3.3

CyberDem0n:
in addition to subj:
* Don't prefetch more than 8 wal files
* Connect from Patroni to postgres via unix_socket
* Set EDITOR=vim


alexeyklyukin:
👍

CyberDem0n:
:+1:

Issue #183, opened 2017-08-29
patroni entered a FATAL state


roll4life:
The following is from fresh install of patroni using the helm chart provided by incubator/patroni. All of the etcd and spilo nodes come up but all of the spilo nodes have the following error. 
Environment is Canonical Distribution of Kubernetes 1.7 running on top of Openstack 

If there is a better place to report this issue, please let me know and I will report elsewhere. 
```
2017-08-29 00:17:19,541 - bootstrapping - INFO - Figuring out my environment (Google? AWS? Local?)
Traceback (most recent call last):
  File "/usr/local/lib/python3.4/dist-packages/requests/packages/urllib3/connection.py", line 141, in _new_conn
    (self.host, self.port), self.timeout, **extra_kw)
  File "/usr/local/lib/python3.4/dist-packages/requests/packages/urllib3/util/connection.py", line 60, in create_connection
    for res in socket.getaddrinfo(host, port, family, socket.SOCK_STREAM):
  File "/usr/lib/python3.4/socket.py", line 533, in getaddrinfo
    for res in _socket.getaddrinfo(host, port, family, type, proto, flags):
socket.gaierror: [Errno -2] Name or service not known

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.4/dist-packages/requests/packages/urllib3/connectionpool.py", line 600, in urlopen
    chunked=chunked)
  File "/usr/local/lib/python3.4/dist-packages/requests/packages/urllib3/connectionpool.py", line 356, in _make_request
    conn.request(method, url, **httplib_request_kw)
  File "/usr/lib/python3.4/http/client.py", line 1125, in request
    self._send_request(method, url, body, headers)
  File "/usr/lib/python3.4/http/client.py", line 1163, in _send_request
    self.endheaders(body)
  File "/usr/lib/python3.4/http/client.py", line 1121, in endheaders
    self._send_output(message_body)
  File "/usr/lib/python3.4/http/client.py", line 951, in _send_output
    self.send(msg)
  File "/usr/lib/python3.4/http/client.py", line 886, in send
    self.connect()
  File "/usr/local/lib/python3.4/dist-packages/requests/packages/urllib3/connection.py", line 166, in connect
    conn = self._new_conn()
  File "/usr/local/lib/python3.4/dist-packages/requests/packages/urllib3/connection.py", line 150, in _new_conn
    self, "Failed to establish a new connection: %s" % e)
[...]

PR #186, opened 2017-09-07
Bugfix/support openstack metadata

roll4life:
refs #183 add initial support for running spilo in kubernetes on openstack

### DIFF
--- a/postgres-appliance/configure_spilo.py
+++ b/postgres-appliance/configure_spilo.py
@@ -217,13 +217,13 @@ def get_provider():
             return PROVIDER_GOOGLE
         else:
             try:
-                r = requests.get('http://instance-data/latest/meta-data/ami-id')  # accessible on AWS, will fail on Openstack
-                if r.ok:
-                    return PROVIDER_AWS
-            except (requests.exceptions.ConnectTimeout, requests.exceptions.ConnectionError):
-                r = requests.get('http://169.254.169.254/latest/meta-data/ami-id')  # is accessible from both AWS and Openstack, Possiblity of misidentification if the guest can't resolve the host instance-data
+                r = requests.get('http://169.254.169.254/openstack/latest/meta_data.json')  # accessible on Openstack, will fail on AWS
                 if r.ok:
                     return PROVIDER_OPENSTACK
+            except Exception:
+                r = requests.get('http://169.254.169.254/latest/meta-data/ami-id')  # is accessible from both AWS and Openstack, Possiblity of misidentification if previous try fails
+                if r.ok:
+                    return PROVIDER_AWS
                 else:
                     return PROVIDER_UNSUPPORTED
     except (requests.exceptions.ConnectTimeout, requests.exceptions.ConnectionError):
--- a/postgres-appliance/configure_spilo.py
+++ b/postgres-appliance/configure_spilo.py
@@ -18,6 +18,7 @@ import requests
 
 PROVIDER_AWS = "aws"
 PROVIDER_GOOGLE = "google"
+PROVIDER_OPENSTACK = "openstack"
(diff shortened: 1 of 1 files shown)


postgres-appliance/configure_spilo.py
```diff
@@ -210,16 +211,21 @@ def deep_update(a, b):
 
 def get_provider():
     try:
-        logging.info("Figuring out my environment (Google? AWS? Local?)")
+        logging.info("Figuring out my environment (Google? AWS? Openstack? Local?)")
         r = requests.get('http://169.254.169.254', timeout=2)
         if r.headers.get('Metadata-Flavor', '') == 'Google':
```
CyberDem0n: I am wondering, may be on Openstack there is something similar available?'
Unique header for example.

roll4life: from what I can tell there is not.  I ran the following to check
```
root@patroni-patroni-0:/home/postgres# python3
Python 3.4.3 (default, Nov 17 2016, 01:08:31) 
[GCC 4.8.4] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import requests
>>> r = requests.get('http://169.254.169.254', timeout=2)
>>> r.headers
{'Date': 'Thu, 07 Sep 2017 22:51:27 GMT', 'Content-Type': 'text/plain; charset=UTF-8', 'Connection': 'keep-alive', 'Content-Length': '98'}
```
I did find that there is an openstack specific path 
`http://169.254.169.254/openstack/latest/meta_data.json`
perhaps this would be a more reliable method of determining that a host is running on openstack.

postgres-appliance/configure_spilo.py
```diff
@@ -210,16 +211,21 @@ def deep_update(a, b):
...
-            r = requests.get('http://169.254.169.254/latest/meta-data/ami-id')  # should be only accessible on AWS
-            if r.ok:
-                return PROVIDER_AWS
-            else:
-                return PROVIDER_UNSUPPORTED
+            try:
+                r = requests.get('http://instance-data/latest/meta-data/ami-id')  # accessible on AWS, will fail on Openstack
+                if r.ok:
+                    return PROVIDER_AWS
+            except (requests.exceptions.ConnectTimeout, requests.exceptions.ConnectionError):
```
CyberDem0n: According to the https://github.com/zalando/spilo/issues/183 it will raise `socket.gaierror: [Errno -2] Name or service not known` (fail to resolve instance-data) on openshift and therefore will to fix the problem

roll4life: @CyberDem0n the problem was actually discovered in plain Kuberenetes running on OpenStack. However, If a user was running OpenShift on OpenStack, they would likely encounter the same error.

CyberDem0n: I understand, but before your changes if was failing in a different place, because `socket.gaierror:` exception hasn't been handled properly.
Now it will fail on the line 220, because DNS can't resolve `instance-data` into ip address.
On the line 223 we are trying to catch only connection error exceptions.
If you do a change:
```python
 -            except (requests.exceptions.ConnectTimeout, requests.exceptions.ConnectionError): 
 +            except Exception: 
```

It will start working, but in addition to that I think it would make sense to change line 224:
```python
-                r = requests.get('http://169.254.169.254/latest/meta-data/ami-id')
+                r = requests.get('http://169.254.169.254/openstack/latest/meta_data.json')
```

roll4life: @CyberDem0n I have implemented the changes you suggested, I removed instance-data and replaced it with the metadata server ip address which should resolve #191

CyberDem0n:
:+1:

avaczi:
:+1:

Issue #188, opened 2017-09-22
Helm chart: cannot login with admin credentials


captjt:
It seems that there isn't a possible way to log into postgres with the "admin" user defined in the Kubernetes helm chart [here](https://github.com/kubernetes/charts/blob/master/incubator/patroni/templates/statefulset-patroni.yaml#L37).

By default here are the users once I log in with `-U postgres` password `tea`.
```
postgres=# \du
                                           List of roles
 Role name  |                         Attributes                         |        Member of
------------+------------------------------------------------------------+-------------------------
 admin      | Create role, Create DB, Cannot login                       | {}
 pgq_admin  | Cannot login                                               | {pgq_reader,pgq_writer}
 pgq_reader | Cannot login                                               | {}
 pgq_writer | Cannot login                                               | {}
 postgres   | Superuser, Create role, Create DB, Replication, Bypass RLS | {}
 robot_zmon | Cannot login                                               | {}
 standby    | Replication                                                | {}
```

It doesn't seem like there is an environment variable for `PGPASSWORD_ADMIN` either in the [docs](https://github.com/zalando/spilo/blob/master/ENVIRONMENT.rst). Anywhere I can help with this issue I'd gladly open a PR as well.

ozgune:
I also ran into this issue when I followed the instructions in https://github.com/kubernetes/charts/tree/master/incubator/patroni

The repo says that the admin password is `cola`. However, when I try to log in as an admin user, I get the following.

<pre>
root@ubuntu:/# psql -U admin -h my-release-patroni.default.svc.cluster.local postgres
Password for user admin: 
psql: FATAL:  password authentication failed for user "admin"
FATAL:  password authentication failed for user "admin"
</pre>

The user `postgres` and password `tea` works.

captjt:
Yeah @ozgune the configurations seem to include the `admin` user, but from what I could find it doesn't seem to actually bootstrap the configurations like [this](https://github.com/zalando/patroni/blob/master/postgres0.yml#L57) would do.

CyberDem0n:
Hi @captjt, hi @ozgune!

[...]

PR #189, opened 2017-09-22
Add admin user for to reflect helm chart

captjt:
Trying to figure out why the helm chart value for `PGPASSWORD_ADMIN` wasn't working. It seems like there wasn't a custom user getting creating in the patroni .yaml file for bootstrap.

Feels like this would fix [https://github.com/zalando/spilo/issues/188](https://github.com/zalando/spilo/issues/188). Let me know if this isn't the case either.

### DIFF
--- a/postgres-appliance/configure_spilo.py
+++ b/postgres-appliance/configure_spilo.py
@@ -148,8 +148,8 @@ bootstrap:
     - hostssl all all 0.0.0.0/0 md5
     - host    all all 0.0.0.0/0 md5
   users:
-    admin:
-      password: '{{PGPASSWORD_ADMIN}}'
+    {{PGUSER_ADMIN}}:
+      password: {{PGPASSWORD_ADMIN}}
       options:
         - createrole
         - createdb
@@ -274,6 +274,7 @@ def get_placeholders(provider):
     placeholders.setdefault('PGDATA', os.path.join(placeholders['PGROOT'], 'pgdata'))
     placeholders.setdefault('PGUSER_STANDBY', 'standby')
     placeholders.setdefault('PGPASSWORD_STANDBY', 'standby')
+    placeholders.setdefault('PGUSER_ADMIN', 'admin')
     placeholders.setdefault('PGPASSWORD_ADMIN', 'cola')
     placeholders.setdefault('PGUSER_SUPERUSER', 'postgres')
     placeholders.setdefault('PGPASSWORD_SUPERUSER', 'zalando')
--- a/postgres-appliance/configure_spilo.py
+++ b/postgres-appliance/configure_spilo.py
@@ -276,7 +276,7 @@ def get_placeholders(provider):
     placeholders.setdefault('PGDATA', os.path.join(placeholders['PGROOT'], 'pgdata'))
     placeholders.setdefault('PGUSER_STANDBY', 'standby')
     placeholders.setdefault('PGPASSWORD_STANDBY', 'standby')
-    placeholders.setdefault('USE_ADMIN', True)
+    placeholders.setdefault('USE_ADMIN', 'PGUSER_ADMIN' in placeholders)
     placeholders.setdefault('PGUSER_ADMIN', 'admin')
     placeholders.setdefault('PGPASSWORD_ADMIN', 'cola')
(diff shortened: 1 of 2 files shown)


postgres-appliance/configure_spilo.py
```diff
@@ -147,6 +147,13 @@ def deep_update(a, b):
   pg_hba:
     - hostssl all all 0.0.0.0/0 md5
     - host    all all 0.0.0.0/0 md5
+  users:
```
CyberDem0n: Can we change it a little bit?
Entire user section should appear only if PGPASSWORD_ADMIN environment variable comes from outside.

postgres-appliance/configure_spilo.py
```diff
@@ -147,6 +147,13 @@ def deep_update(a, b):
...
+    {{PGUSER_ADMIN}}:
+      password: {{PGPASSWORD_ADMIN}}
+      options:
+        - createrole
+        - createdb
+        - login
```
CyberDem0n: Patroni is using "CREATE USER" which grants login option by default.

postgres-appliance/configure_spilo.py
captjt: Yeah I can do that - good point 👍

captjt:
I actually don't think doing something like `{{#...?}}` is a good option. Would rather have a `true/false` variable for those. What do you think? @CyberDem0n -- this would mean we need to add something to the helm chart though.

postgres-appliance/configure_spilo.py
```diff
@@ -268,6 +276,9 @@ def get_placeholders(provider):
     placeholders.setdefault('PGDATA', os.path.join(placeholders['PGROOT'], 'pgdata'))
     placeholders.setdefault('PGUSER_STANDBY', 'standby')
     placeholders.setdefault('PGPASSWORD_STANDBY', 'standby')
+    placeholders.setdefault('USE_ADMIN', True)
```
CyberDem0n: We don't really need to pass 'USE_ADMIN' from outside, but we can calculate its value from the presence of 'PGPASSWORD_ADMIN' in environment:
```diff
-    placeholders.setdefault('USE_ADMIN', True)
+    placeholders.setdefault('USE_ADMIN', 'PGPASSWORD_ADMIN' in placeholders)
```

If you wish, it would be still possible to pass it.

captjt: Nice, I like that as well. So it can be overridden if you set `USE_ADMIN` but if you don't it will default to whether or not `PGPASSWORD_ADMIN` or whatever we define is actually there? Sorry first time using mustache templating.

postgres-appliance/configure_spilo.py
```diff
@@ -276,7 +276,7 @@ def get_placeholders(provider):
     placeholders.setdefault('PGDATA', os.path.join(placeholders['PGROOT'], 'pgdata'))
     placeholders.setdefault('PGUSER_STANDBY', 'standby')
     placeholders.setdefault('PGPASSWORD_STANDBY', 'standby')
-    placeholders.setdefault('USE_ADMIN', True)
+    placeholders.setdefault('USE_ADMIN', 'PGUSER_ADMIN' in placeholders)
```
CyberDem0n: Actually I was talking about `PGPASSWORD_ADMIN`.

Admin user name already has sane default value and can be skipped from environment, but password you will probably override.

captjt: Very true -- should I keep a defaulted `PGUSER_ADMIN` and not make it overridable?

CyberDem0n: `placeholders.setdefault('PGUSER_ADMIN', 'admin )` -- this one just leave as it is.
If somebody will define a `PGUSER_ADMIN` environment variable with a different value, this value will override `admin`.

captjt: Sorry about all the delays @CyberDem0n -- should be fixed now to check for the defined `PGPASSWORD_ADMIN` placeholder.

CyberDem0n:
:+1:

avaczi:
:+1:

CyberDem0n:
Merged, thank you @captjt

PR #192, opened 2017-09-27
Request AWS instance data via IP

### LINKED ISSUE #191: AWS support uses unsupported DNS name
While troubleshooting a chart that depends on patroni, we discovered that Spilo tries to get AWS instance data via a request to http://instance-data/... in one location only (other locations use the IP address 169.254.169.254). This hostname isn't documented or supported by AWS:
https://forums.aws.amazon.com/message.jspa?messageID=536813
and in some cases it doesn't work (including ours). We were able to work around by cramming it into our kube DNS server but it would be more consistent to make the request by IP address.

One other note: while reading the current issue list, I noticed https://github.com/zalando/spilo/pull/186 and read through the comments. I'm not familiar with how OpenStack is supported but it sounded like they might be relying on the DNS name lookup failure? I'm not sure how fixing this might affect that behavior.

fastjames:
AWS recommends fetching instance data via IP rather than hostname.

source: https://forums.aws.amazon.com/message.jspa?messageID=536813

Fixes #191

### DIFF
--- a/postgres-appliance/configure_spilo.py
+++ b/postgres-appliance/configure_spilo.py
@@ -241,7 +241,7 @@ def get_instance_metadata(provider):
         if not USE_KUBERNETES:
             mapping.update({'id': 'id'})
     elif provider == PROVIDER_AWS:
-        url = 'http://instance-data/latest/meta-data'
+        url = 'http://169.254.169.254/latest/meta-data'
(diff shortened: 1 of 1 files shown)


CyberDem0n:
I've just merged https://github.com/zalando/spilo/pull/186 which already contains this fix.

PR #203, opened 2018-01-10
Improve support for namespaces

alexeyklyukin:
if a NAMESPACE parameter is passed to Spilo and is not set to the value 'default', use it as a DCS namespace. In addition, in case of the Kubernetes being set as a DCS, read this parameter from the
POD_NAMESPACE, unless it it set explicitely.

Allow setting the prefix and the suffix for the cluster-specific folder inside the WAL bucket. If NAMESPACE parameter is set to a non-default value, the folder prefix is set to the NAMESPACE value, followed by a dash, unless the folder prefix is set explicitely.


CyberDem0n:
:+1:

a1exsh:
:+1:

PR #205, opened 2018-01-19
Patroni 1.4.1

CyberDem0n:
In addition to that use a Patroni configfile for patronictl. It contains cluster name and therefore makes optional specifying it for some of the patronictl commands like `list`, `edit-config` `pause` and so on.


CyberDem0n:
:+1:

a1exsh:
:+1:

CyberDem0n:
:+1:

PR #222, opened 2018-04-11
Configure postgres version via PGVERSION env variable

CyberDem0n:
before that it was possible to pass postgresq.bin_dir in the
SPILO_CONFIGURATION:

```yaml
postgresql:
  bin_dir: /usr/lib/postgresql/9.5/bin
```

The old method will still work and has a priority over PGVERSION.


alexeyklyukin:
👍

PR #224, opened 2018-04-26
Define the failed auth views

avaczi:
The aim is to give a means to `robot_zmon` to check failed logins, without allowing access to other log lines.


Jan-M:
:+1:

PR #226, opened 2018-05-07
Migrate to ubuntu:18.04

CyberDem0n:
Since it already has patroni package we will install most of the dependencies as ubuntu packages.
Unfortunately some of the postgres packages must be rebuilt from sources, because they exist only for postgresql-10


postgres-appliance/Dockerfile.build
```diff
@@ -71,126 +69,129 @@ ENV PGVERSION="$PGVERSION" \
...
+                && if [ $version != "9.3" ]; then \
+                    EXTRAS="$EXTRAS postgresql-${version}-amcheck postgresql-${version}-pglogical postgresql-${version}-wal2json" \
+                    # postgresql-${version}-partman -- rebuild packages
+                    && if [ $version != "9.4" ]; then \
+                        EXTRAS="$EXTRAS postgresql-${version}-pgl-ddl-deploy"; \
+                    fi; \
                 fi; \
             fi \
 
+#            if [ $version != "9.3" ] && [ $version != "9.4" ]; then \
```
zerg-junior: Why these lines are commented out ?

zerg-junior:
:+1:

CyberDem0n:
:+1:

PR #230, opened 2018-05-15
Enable customization of Patroni Kubernetes labels.

alexeyklyukin:
Add environment variables to customize names and values of labels vused
by Patroni on Kubernets to locate its metadata. At least one of those
labels (cluster name) has different default value between Spilo
('version'), Patroni and the postgres-operator ('cluster-name').  This
change allows to pass the operator label names directly to Spilo (and
Patroni), w/o relying on defaults to match.


CyberDem0n:
:+1:

PR #234, opened 2018-05-28
Set WALE_LOG_DESTINATION to stderr

CyberDem0n:
Otherwise we get a lot of exceptions, because /dev/log doesn't exist.


Jan-M:
:+1:

PR #235, opened 2018-05-29
Security fix

CyberDem0n:
Protect user management functions from CVE-2018-1058.

Some of the functions created with security definer are owned by superuser, therefore it is very important to make sure that they are executed with search_path = 'pg_catalog'


zerg-junior:
for reference: [ Guide to CVE-2018-1058: Protect Your Search Path](https://wiki.postgresql.org/wiki/A_Guide_to_CVE-2018-1058:_Protect_Your_Search_Path)

zerg-junior:
:+1:

PR #236, opened 2018-05-29
Make post_init.sh script idempotent

CyberDem0n:
The second run will not produce errors but rather will fix some discrepancies between current and desired state.
In addition to that it makes zmon_utils compatible with postgres >= 10.

On the next step we can call this script after every "promote" event. It will eliminate the need to manually deploy future changes.

### DIFF
--- a/postgres-appliance/_zmon_schema.dump
+++ b/postgres-appliance/_zmon_schema.dump
@@ -1,9 +1,24 @@
 RESET ROLE;
 SET ROLE TO postgres;
 
-CREATE LANGUAGE plpython3u;
-
 DROP SCHEMA IF EXISTS zmon_utils CASCADE;
+
+DO $$
+BEGIN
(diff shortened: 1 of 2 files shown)


postgres-appliance/_zmon_schema.dump
```diff
@@ -1,9 +1,24 @@
...
+    PERFORM * FROM pg_catalog.pg_language WHERE lanname = 'plpython3u' AND NOT EXISTS (SELECT 1 FROM pg_catalog.pg_extension WHERE extname = 'plpython3u');
+    IF FOUND THEN
+        CREATE EXTENSION plpython3u FROM UNPACKAGED;
+    ELSE
+        CREATE EXTENSION IF NOT EXISTS plpython3u;
+    END IF;
+END;$$;
+
+-- remove plpython2 leftovers
+DROP LANGUAGE IF EXISTS plpythonu;
```
mgomezch: Is plpython2 currently usable?

mgomezch:
:+1:

CyberDem0n:
:+1:

PR #238, opened 2018-05-29
Call post_init.sh on promote

CyberDem0n:
This way we can make sure that the recent version of user_management and monitoring functions are deployed.


alexeyklyukin:
👍

mgomezch:
:+1:

PR #241, opened 2018-06-07
Don't do backup right after failover

CyberDem0n:
there were two problems:
1. backup script called from the launch.sh didn't honor `BACKUP_NUM_TO_RETAIN` and therefore using default value = 2
2. honoring of `BACKUP_NUM_TO_RETAIN` doesn't really help us when we have a lot of switchovers/failovers due to rolling upgrades on k8s, because it will result in all backups made just recently.


postgres-appliance/scripts/postgres_backup.sh
```diff
@@ -10,23 +10,34 @@ function log
...
-envdir "${WALE_ENV_DIR}" wal-e --aws-instance-profile delete --confirm retain "${NUM_TO_RETAIN}"
+readonly WAL_E="envdir $WALE_ENV_DIR wal-e --aws-instance-profile"
+
+BEFORE=""
+
+readonly NOW=$(date +%s -u)
+while read name last_modified rest; do
+    last_modified=$(date +%s -ud "$last_modified")
+    if [ $(((NOW-last_modified)/86400)) -gt $NUM_TO_RETAIN ]; then
+        if [ -z "$BEFORE" ] || [ "$last_modified" -lt "$BEFORE_TIME" ]; then
```
alexeyklyukin: we should look for the earliest backup that is within NUM_TO_RETAIN days from now (to remove everything before that)

avaczi: having another look, maybe using a better name instead of `NUM_TO_RETAIN` would make sense with this change.

avaczi:
:+1:

alexeyklyukin:
👍

CyberDem0n:
:+1:

PR #257, opened 2018-08-31
Adjust log shipping schedule to match the value stated in the docs

zerg-junior:
this value should have been set during the initial work on the PR, but I seemingly only mentioned the correct default i[n the docs](https://github.com/zalando/spilo/blob/master/ENVIRONMENT.rst)

### DIFF
--- a/postgres-appliance/scripts/configure_spilo.py
+++ b/postgres-appliance/scripts/configure_spilo.py
@@ -385,7 +385,7 @@ def get_placeholders(provider):
     placeholders.setdefault('CLONE_TARGET_TIME', '')
     placeholders.setdefault('CLONE_TARGET_INCLUSIVE', True)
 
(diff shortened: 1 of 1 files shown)


CyberDem0n:
:+1:

postgres-appliance/scripts/configure_spilo.py
```diff
@@ -385,7 +385,7 @@ def get_placeholders(provider):
     placeholders.setdefault('CLONE_TARGET_TIME', '')
     placeholders.setdefault('CLONE_TARGET_INCLUSIVE', True)
 
-    placeholders.setdefault('LOG_SHIP_SCHEDULE', '1 0 * * *')
+    placeholders.setdefault('LOG_SHIP_SCHEDULE', '00 02 * * *')
```
CyberDem0n: The old value looks correct for me, we want to push logs right after its rotated at midnight.

PR #258, opened 2018-09-06
Add wal-g support

CyberDem0n:
To enable it one has to set `USE_WALG` env to `true`.
Most of the wal-g [configuration](https://github.com/wal-g/wal-g#configuration) options are transparently propagated.
Besides that, one can define:
* `WALG_BACKUP_COMPRESSION_METHOD` to override compression method for backups.
* `WALG_BACKUP_FROM_REPLICA=true` -- to take backups from the replica. You shouldn't use it if there is more than one replica.


postgres-appliance/scripts/configure_spilo.py
```diff
@@ -417,7 +418,12 @@ def get_placeholders(provider):
             else:
                 placeholders['CALLBACK_SCRIPT'] = 'patroni_aws'
 
-    placeholders['USE_WALE'] = bool(placeholders.get('WAL_S3_BUCKET') or placeholders.get('WAL_GCS_BUCKET') or placeholders.get('WAL_SWIFT_BUCKET'))
+    use_walg = str(placeholders['USE_WALG']).lower() == 'true' and bool(placeholders.get('WAL_S3_BUCKET'))
+    placeholders['USE_WALE'] = bool(use_walg or placeholders.get('WAL_GCS_BUCKET')
```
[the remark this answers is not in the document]
CyberDem0n: wal-g is a kind of subset of wal-e, because it supports only S3.

postgres-appliance/scripts/configure_spilo.py
```diff
@@ -627,11 +625,14 @@ def write_crontab(placeholders, overwrite):
                 return logging.warning('Cron is already configured. (Use option --force to overwrite cron)')
 
     lines = ['PATH={PATH}'.format(**placeholders)]
-    lines += ['{BACKUP_SCHEDULE} /scripts/postgres_backup.sh "{WALE_ENV_DIR}" "{PGDATA}" "{BACKUP_NUM_TO_RETAIN}"'
-              .format(**placeholders)]
+
+    if bool(placeholders.get('USE_WALE')):
```
zerg-junior: one of the changes to crontab seems to lead to:

> 2018-09-06 11:51:06,274 - bootstrapping - INFO - Configuring crontab                                                   
"-":1: bad minute                                                                                         
errors in crontab file, can't install.

for example when the image was run as:
```bash
./build.sh --build-arg COMPRESS=true --tag spilo-test-walg .
docker run -it -e LOG_S3_BUCKET="test-bucket"  spilo-test-walg:latest
```
with 
```bash
docker run -it -e WAL_S3_BUCKET="test-bucket"  spilo-test-walg:latest
```
this error is not logged, but the crontab still seems to be empty for the user `postgres`

postgres-appliance/scripts/configure_spilo.py
zerg-junior: does it mean that from now on we can stream to S3 using only `wal-g` ?  Also, when I try 
```bash
 docker run -it -e WAL_S3_BUCKET="test-bucket"  spilo-test-walg:latest
```

the WALE_ENV_DIR is not written

PR #259, opened 2018-09-06
Fix bugs in configure_spilo and postgres_backup

CyberDem0n:
* string formatting has higher precedence than concatenation
* postgres_backup.sh now need only one mandatory argument


erthalion:
:+1:

PR #267, opened 2018-10-24
fix docker build error 

tnextday:
This pr fixes the following build errors:

1. error when using the latest ubuntu:18.04 image
https://github.com/tianon/docker-brew-ubuntu-core/pull/135 , the universe deb-src is not enabled by default
2. error when PGOLDVERSIONS is set to ""
3. rebuild the pg extension package only for the installed version

In addition, there is a patch for python-etcd v0.4.3, https://github.com/jplana/python-etcd/pull/196, can I submit it to this pr?


tnextday:
I create another pr https://github.com/zalando/spilo/pull/269

PR #278, opened 2018-12-06
Make it possible to use wal-g only for restore

CyberDem0n:
This commit introduces two new env variables:
1. USE_WAL_BACKUP
2. USE_WAL_RESTORE

USE_WAL=true will set both aforementioned variables to true if they are not defined.


postgres-appliance/scripts/configure_spilo.py
```diff
@@ -372,6 +372,8 @@ def get_placeholders(provider):
     placeholders.setdefault('WALE_ENV_DIR', os.path.join(placeholders['PGHOME'], 'etc', 'wal-e.d', 'env'))
     placeholders.setdefault('USE_WALE', False)
     placeholders.setdefault('USE_WALG', False)
+    placeholders.setdefault('USE_WALG_BACKUP', placeholders['USE_WALG'])
```
zerg-junior: Briefly documenting this in the list of env vars will be of worth for other users

PR #281, opened 2019-01-02
Update timescaledb extension after a failover if it is necessary

CyberDem0n:
We don't install old binaries of timescaledb into the image and therefore right after failover timescaledb extension needs to be explicitly updated.

In addition to that, bump versions of wal-g, plpgsql_check and timescaledb and install plprofiler from the new repository.


Jan-M:
:+1:

PR #282, opened 2019-01-04
Execute pg_upgrade after custom bootstrap if major version doesn't match

CyberDem0n:
Before doing the actual call of `pg_upgrade`, the `maybe_pg_upgrade.py` script starts the old postgres binary, waits until `pg_is_in_recovery()` starts returning `False` and does a clean shutdown.

### DIFF
--- a/postgres-appliance/bootstrap/clone_with_wale.py
+++ b/postgres-appliance/bootstrap/clone_with_wale.py
@@ -85,10 +85,11 @@ def run_clone_from_s3(options):
 def main():
     options = read_configuration()
     try:
-        return run_clone_from_s3(options) and call_maybe_pg_upgrade()
+        run_clone_from_s3(options)
     except Exception:
         logger.exception("Clone failed")
         return 1
+    return call_maybe_pg_upgrade()
 
 
 if __name__ == '__main__':
--- a/postgres-appliance/bootstrap/clone_with_wale.py
+++ b/postgres-appliance/bootstrap/clone_with_wale.py
@@ -5,6 +5,9 @@ import csv
 import logging
 import os
 import subprocess
+import sys
+
+from maybe_pg_upgrade import call_maybe_pg_upgrade
 
 from collections import namedtuple
 from dateutil.parser import parse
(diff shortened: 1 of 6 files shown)


postgres-appliance/bootstrap/clone_with_wale.py
```diff
@@ -82,11 +85,11 @@ def run_clone_from_s3(options):
 def main():
     options = read_configuration()
     try:
-        return run_clone_from_s3(options)
+        return run_clone_from_s3(options) and call_maybe_pg_upgrade()
```
Jan-M: seeing this twice, any way this can be merged?

and the exception log line could then also hin on "upgrade failed".

CyberDem0n: There is no "twice".

Error messaging indeed could be improved.

Jan-M: Could in general need some more logging? The upgrade Python code looks very empty.

postgres-appliance/bootstrap/pg_upgrade.py
```diff
@@ -0,0 +1,58 @@
...
+                           '-b', self._old_bin_dir, '-B', self._bin_dir,
+                           '-d', self._old_data_dir, '-D', self._data_dir]
+        if 'username' in self._superuser:
+            pg_upgrade_args += ['-U', self._superuser['username']]
+
+        return subprocess.call([self._pgcommand('pg_upgrade')] + pg_upgrade_args) == 0
+
+    def do_upgrade(self, version, initdb_config):
+        self._data_dir = os.path.abspath(self._data_dir)
+        self._old_data_dir = self._data_dir + 'old'
```
Jan-M: No + "_"?

postgres-appliance/bootstrap/maybe_pg_upgrade.py
```diff
@@ -0,0 +1,84 @@
...
+    my_name = os.path.abspath(inspect.getfile(inspect.currentframe()))
+    ret = subprocess.call([sys.executable, my_name, os.path.join(os.getenv('PGHOME'), 'postgres.yml')])
+    if ret != 0:
+        logger.error('%s script failed', my_name)
+    return ret
+
+
+if __name__ == '__main__':
+    logging.basicConfig(format='%(asctime)s maybe_pg_upgrade %(levelname)s: %(message)s', level='INFO')
+    if len(sys.argv) > 5 and sys.argv[1] == 'pg_ctl_start':
```
zerg-junior: what is so special about this number of arguments ?

CyberDem0n: When starting postgres we expect to get at least some amount of arguments:
`script.py pg_ctl_start postgres -D pgdata ...`

Issue #265, opened 2018-09-27
K8s template with Kubernetes as DCS not working


kmoppel:
Hi guys!

Not sure if my own config issue (new to Kubernetes) or a real issue but the provided [template](
https://github.com/zalando/spilo/blob/master/kubernetes/spilo_kubernetes_aws.yaml) leaves the standby in failed state with quite a dry error message:

2018-09-27 07:41:09,233 INFO: Lock owner: zalandodemo01-0; I am zalandodemo01-1
2018-09-27 07:41:09,247 INFO: trying to bootstrap from leader 'zalandodemo01-0'
2018-09-27 07:41:09,248 ERROR: failed to bootstrap from leader 'zalandodemo01-0'
2018-09-27 07:41:09,248 INFO: Removing data directory: /home/postgres/pgdata/pgroot/data

They adjustments I made to the template:
Removed params - ETCD_HOST, WAL_S3_BUCKE, LOG_S3_BUCKET, BACKUP_SCHEDULE
Added params - DCS_ENABLE_KUBERNETES_API='true'
image: registry.opensource.zalan.do/acid/spilo-10:1.4-p16
serviceAccountName: default (gave cluster-admin priviliges beforehand)

Got the same error also in the KUBERNETES_USE_CONFIGMAPS mode.

And actually to get even so far had to give a new name to the declared Endpoint - otherwise it was erroring out with "HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"endpoints \"zalandodemo01\" already exists","reason":"AlreadyExists","details":{"name":"zalandodemo01","kind":"endpoints"},"code":409}"

Used K8s: v1.10 on Minikube (tried also with GKE actually)

Any ideas / hints would be very welcome! And btw, is there a way to add some verbosity to the logs?

CyberDem0n:
Hi @kmoppel,

this template seems to be very outdated, people usually use [helm chart](https://github.com/helm/charts/tree/master/incubator/patroni) to deploy spilo on k8s.

I'll try to prepare the new template.

One can add `PATRONI_LOGLEVEL=DEBUG` environment variable before starting Patroni and get more logs, but most of them are just communication between Patroni and DCS.

You can also look into Pod logs and postgres from the master, they should definitely have something.

kmoppel:
[...]

PR #293, opened 2019-01-30
Update spilo_kubernetes.yaml manifest

CyberDem0n:
It makes possible to run spilo on k8s without the Patroni helm chart or the postgres-operator. There is one minor difference though, helm chart, postgres-operator, and spilo_kubernetes.yaml are using different label names. The cluster name (and names of any other objects) is hardcoded to the `zalandodemo01`.

Another very important point is, the manifest is configured to use kubernetes API and endpoints for leader elections, i.e. you don't need to deploy etcd cluster inside the kubernetes.

In addition to that, manifest contains the "operator" service account, which is granted to do all necessary permissions on k8s objects.

All the aforementioned facts make possible to deploy the manifest on `minikube` by just running a single command: `kubectl create -f spilo_kubernetes.yaml`

Fixes https://github.com/zalando/spilo/issues/265

### DIFF
--- a/kubernetes/spilo_kubernetes.yaml
+++ b/kubernetes/spilo_kubernetes.yaml
@@ -1,7 +1,7 @@
 apiVersion: apps/v1beta1
 kind: StatefulSet
 metadata:
-  name: &cluster_name zalandodemo01 
+  name: &cluster_name zalandodemo01
   labels:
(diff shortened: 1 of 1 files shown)


Jan-M:
:+1:

Issue #295, opened 2019-02-01
Missing pgpass file configuration


farhan5900:
There is a configuration **pgpass** in Patroni to specify the location of a password file. This configuration is missing in Spilo. By default, in Patroni it is taken as home directory of current user appended with file name (https://github.com/zalando/patroni/blob/10bbf0c3c545421c8f247fb022bee0e56bf9813d/patroni/postgresql.py#L123). Patroni allows it to be set by exporting 'PATRONI_POSTGRESQL_PGPASS', but when we run Spilo, we are unable to set it.

PR #296, opened 2019-02-01
added PGPASS_FILE environment

farhan5900:
It resolves issue #295. Some minor fix in ENVIRONMENT.rst also.


CyberDem0n:
I don't really like that you change the default behavior. Somebody might decide to put a custom `.pgpass` into the image and Patroni will silently overwrite it.

PR #303, opened 2019-02-13
Update extensions

CyberDem0n:
* bump bg_mon commitid
* bump timescaledb version
* add and enable pg_stat_kcache extension


postgres-appliance/Dockerfile.build
```diff
@@ -225,7 +228,7 @@ RUN export DEBIAN_FRONTEND=noninteractive \
     fi \
 
     # build and install missing packages
-    && apt-get install -y postgresql-server-dev-all postgresql-server-dev-${PGVERSION} \
+    && apt-get install -y postgresql-server-dev-all postgresql-server-dev-9.3 \
```
Jan-M: why is this? 9.3?

CyberDem0n: 9.3 is EOL and isn't included as a dependency in postgresql-server-dev-all.

PR #304, opened 2019-02-14
Introduce metric_helpers schema

CyberDem0n:
it contains three function/view with table/btree bloat queries and wrapper on pg_stat_statements.

In order to avoid potential problems with pg_upgrade, the upgrade script will drop the metric_helpers schema before the upgrade.

### DIFF
--- a/postgres-appliance/bootstrap/pg_upgrade.py
+++ b/postgres-appliance/bootstrap/pg_upgrade.py
@@ -41,7 +41,7 @@ class PostgresqlUpgrade(Postgresql):
             conn_kwargs['database'] = d[0]
             with self._get_connection_cursor(**conn_kwargs) as cur:
                 cur.execute("SET synchronous_commit = 'local'")
-                cur.execute("DROP SCHEMA IF EXISTS metric_helpers CASCADE")
+                cur.execute("DROP FUNCTION metric_helpers.pg_stat_statements(boolean) CASCADE")
 
     def pg_upgrade(self):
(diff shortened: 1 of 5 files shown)


postgres-appliance/bootstrap/pg_upgrade.py
```diff
@@ -28,6 +31,18 @@ def copy_configs(self):
...
+        conn_kwargs = self._local_connect_kwargs
+        for p in ['connect_timeout', 'options']:
+            conn_kwargs.pop(p, None)
+
+        for d in self.query('SELECT datname FROM pg_catalog.pg_database WHERE datallowconn'):
+            logger.info('Executing "DROP SCHEMA IF EXISTS metric_helpers" in the database="%s"', d[0])
+            conn_kwargs['database'] = d[0]
+            with self._get_connection_cursor(**conn_kwargs) as cur:
+                cur.execute("SET synchronous_commit = 'local'")
+                cur.execute("DROP SCHEMA IF EXISTS metric_helpers CASCADE")
```
Jan-M: Can we do this without dropping the schema?

Jan-M:
Did we not also want to disable tracking of utility statements?

postgres-appliance/bootstrap/pg_upgrade.py
CyberDem0n: In theory it should be enough to drop function and view, but dropping the entire schema is the safest. post_init.sh will restore it after the upgrade.

PR #305, opened 2019-02-15
Control AWS instance profile behaviour from the envdir

CyberDem0n:
configure_spilo.py will write 'true' into AWS_INSTANCE_PROFILE file in the envdir when S3 is requested and AWS_SECRET_ACCESS_KEY/AWS_ACCESS_KEY_ID are not specified.

Hope it will resolve https://github.com/zalando-incubator/postgres-operator/issues/488


Jan-M:
:+1:

PR #306, opened 2019-02-17
Added logic to remove circular links

mpritter76:
Attempted to do a build for postgresql version 10 and realized it was creating circular links.

This ultimately was causing the database to become unable to bootstrap.

I don't totally understand the reasoning behind the sections of the code that create links back to the postgresql 9.5 version but without the changes I've added it effectively created circular links and wiped out the actual files. 

When I ran the build to produce the squashed image these were the version settings in the Dockerfile.build file:
```
ARG PGOLDVERSIONS="9.3 9.4 9.5 9.6"
ARG PGVERSION="10"
```

Example of one file (topology.sql) that ends up with a circular link  
```
# pwd
/usr/share/postgresql/9.5/contrib/postgis-2.5
# ls -l /usr/share/postgresql/10/contrib/postgis-2.5/topology.sql
lrwxrwxrwx. 1 root root 58 Feb 17 18:58 /usr/share/postgresql/10/contrib/postgis-2.5/topology.sql -> /usr/share/postgresql/9.5/contrib/postgis-2.5/topology.sql
# ls -l /usr/share/postgresql/9.5/contrib/postgis-2.5/topology.sql
lrwxrwxrwx. 1 root root 57 Feb 17 18:58 /usr/share/postgresql/9.5/contrib/postgis-2.5/topology.sql -> /usr/share/postgresql/10/contrib/postgis-2.5/topology.sql
```

Their might be a better way to solve this issue like removing the version PGVERSION from the list of versions it's cycling through when creating the link but it's probably a good check to not use links as a source for a ln your creating.

Anyway, feel free to share your thoughts on this. I'm really curious about the reasoning behind the 9.5 links.

### DIFF
--- a/postgres-appliance/Dockerfile.build
+++ b/postgres-appliance/Dockerfile.build
@@ -279,6 +279,9 @@ RUN export DEBIAN_FRONTEND=noninteractive \
         && if [ -d /usr/share/postgresql/9.5/contrib/postgis-$POSTGIS_VERSION ]; then \
             cd /usr/share/postgresql/9.5/contrib/postgis-$POSTGIS_VERSION \
             && for f in *.sql *.pl; do \
+                 if [ -L /usr/share/postgresql/9.5/contrib/postgis-$POSTGIS_VERSION/$f ]; then \
+                   continue; \
+                 fi; \
                 for v in 9.4 9.6 10; do \
                     if [ -f /usr/share/postgresql/$v/contrib/postgis-$POSTGIS_VERSION/$f ] \
                             && [ ! -L /usr/share/postgresql/$v/contrib/postgis-$POSTGIS_VERSION/$f ] \
@@ -328,6 +331,9 @@ RUN export DEBIAN_FRONTEND=noninteractive \
         && if [ -d /usr/share/postgresql/9.5/extension ]; then \
             cd /usr/share/postgresql/9.5/extension \
             && for f in *.sql *.control; do \
+                  if [ -L /usr/share/postgresql/9.5/extension/$f ]; then \
+                     continue; \
+                  fi; \
                 for v in 9.3 9.4 9.6 10; do \
                     if [ -f /usr/share/postgresql/$v/extension/$f ] \
                             && [ ! -L /usr/share/postgresql/$v/extension/$f ] \
--- a/postgres-appliance/Dockerfile.build
+++ b/postgres-appliance/Dockerfile.build
@@ -282,7 +282,10 @@ RUN export DEBIAN_FRONTEND=noninteractive \
                  if [ -L /usr/share/postgresql/9.5/contrib/postgis-$POSTGIS_VERSION/$f ]; then \
                    continue; \
                  fi; \
-                for v in 9.4 9.6 10; do \
+                 for v in $PGOLDVERSIONS; do \
+                     if [ $v == 9.5 ]; then \
+                       continue; \
+                     fi; \
                     if [ -f /usr/share/postgresql/$v/contrib/postgis-$POSTGIS_VERSION/$f ] \
                             && [ ! -L /usr/share/postgresql/$v/contrib/postgis-$POSTGIS_VERSION/$f ] \
                             && diff /usr/share/postgresql/$v/contrib/postgis-$POSTGIS_VERSION/$f $f > /dev/null; then \
@@ -334,7 +337,10 @@ RUN export DEBIAN_FRONTEND=noninteractive \
                   if [ -L /usr/share/postgresql/9.5/extension/$f ]; then \
                      continue; \
                   fi; \
-                for v in 9.3 9.4 9.6 10; do \
+                  for v in $PGOLDVERSIONS; do \
+                      if [ $v == 9.5 ]; then \
+                        continue; \
+                      fi; \
                     if [ -f /usr/share/postgresql/$v/extension/$f ] \
                             && [ ! -L /usr/share/postgresql/$v/extension/$f ] \
                             && diff /usr/share/postgresql/$v/extension/$f $f > /dev/null; then \
--- a/postgres-appliance/Dockerfile.build
+++ b/postgres-appliance/Dockerfile.build
@@ -279,14 +279,11 @@ RUN export DEBIAN_FRONTEND=noninteractive \
         && if [ -d /usr/share/postgresql/9.5/contrib/postgis-$POSTGIS_VERSION ]; then \
             cd /usr/share/postgresql/9.5/contrib/postgis-$POSTGIS_VERSION \
             && for f in *.sql *.pl; do \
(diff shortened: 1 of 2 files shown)


CyberDem0n:
Hi @mpritter76,

Indeed, if you change PGVERSION from 11 to 10 you'll get such weird behavior because I didn't really expect that somebody would like to build the image without pg11.
What is your goal? Why do you want to get rid of pg11? Do you want to make the image smaller or you don't know how to use not the latest postgres version when you start container?
If it a second one, you have even to options:
1. Specify [PGVERSION](https://github.com/zalando/spilo/blob/master/postgres-appliance/scripts/configure_spilo.py#L750-L755)=10 environment variable when starting a container.
2. Specify the `postgresql.bin_dir` in the `SPILO_CONFIGURATION` environment variable (like below)
```
SPILO_CONFIGURATION='{"postgresql": {"bin_dir": "/usr/lib/postgresql/10/bin"}}'
```

`SPILO_CONFIGURATION` is very powerful, you can control nearly everything with it. When it is supplied, [configure_spilo.py](https://github.com/zalando/spilo/blob/master/postgres-appliance/scripts/configure_spilo.py#L743-L748) will use it for overriding the default configuration generated by the script.

mpritter76:
You are correct, I didn't not understand to bootstrap a lower version I simply needed to set an environment variable in the container.   Was that in the documentation? I must have totally missed that. 

To be clear, do I need to set both variables? Or just the PGVERSION? 

Regarding the PR, I think it can still be employed...  It probably could still be employed. It would eliminate the risk of using links as the source for other links and cycling through the 'PGOLDVERSIONS' stills gives you the benefit of listing the variables in a single place. Up to you at this point.

Thanks again for the timely response!

postgres-appliance/Dockerfile.build
```diff
@@ -279,7 +279,13 @@ RUN export DEBIAN_FRONTEND=noninteractive \
         && if [ -d /usr/share/postgresql/9.5/contrib/postgis-$POSTGIS_VERSION ]; then \
             cd /usr/share/postgresql/9.5/contrib/postgis-$POSTGIS_VERSION \
             && for f in *.sql *.pl; do \
-                for v in 9.4 9.6 10; do \
+                 if [ -L /usr/share/postgresql/9.5/contrib/postgis-$POSTGIS_VERSION/$f ]; then \
```
CyberDem0n: `if [ -L /usr/share/postgresql/9.5/contrib/postgis-$POSTGIS_VERSION/$f ]; then \` could be changed to `if [ -L $f ]; then \`, we are already in the `/usr/share/postgresql/9.5/contrib/postgis-$POSTGIS_VERSION` directory.

postgres-appliance/Dockerfile.build
```diff
@@ -328,7 +334,13 @@ RUN export DEBIAN_FRONTEND=noninteractive \
         && if [ -d /usr/share/postgresql/9.5/extension ]; then \
             cd /usr/share/postgresql/9.5/extension \
             && for f in *.sql *.control; do \
-                for v in 9.3 9.4 9.6 10; do \
+                  if [ -L /usr/share/postgresql/9.5/extension/$f ]; then \
```
CyberDem0n: `if [ -L $f ]; then \`

postgres-appliance/Dockerfile.build
```diff
@@ -279,7 +279,13 @@ RUN export DEBIAN_FRONTEND=noninteractive \
         && if [ -d /usr/share/postgresql/9.5/contrib/postgis-$POSTGIS_VERSION ]; then \
             cd /usr/share/postgresql/9.5/contrib/postgis-$POSTGIS_VERSION \
             && for f in *.sql *.pl; do \
-                for v in 9.4 9.6 10; do \
+                 if [ -L /usr/share/postgresql/9.5/contrib/postgis-$POSTGIS_VERSION/$f ]; then \
+                   continue; \
+                 fi; \
+                 for v in $PGOLDVERSIONS; do \
+                     if [ $v == 9.5 ]; then \
```
CyberDem0n: I think it would be better to combine it with the next if block:
`if [ $v != 9.5 ] && [ -f /usr/share/postgresql/$v/contrib/postgis-$POSTGIS_VERSION/$f ] \`

postgres-appliance/Dockerfile.build
```diff
@@ -328,7 +334,13 @@ RUN export DEBIAN_FRONTEND=noninteractive \
         && if [ -d /usr/share/postgresql/9.5/extension ]; then \
             cd /usr/share/postgresql/9.5/extension \
             && for f in *.sql *.control; do \
-                for v in 9.3 9.4 9.6 10; do \
+                  if [ -L /usr/share/postgresql/9.5/extension/$f ]; then \
+                     continue; \
+                  fi; \
+                  for v in $PGOLDVERSIONS; do \
+                      if [ $v == 9.5 ]; then \
```
CyberDem0n: Let's combine it with the next if block:
`if [ $v != 9.5 ] && [ -f /usr/share/postgresql/$v/extension/$f ] \`

postgres-appliance/Dockerfile.build
```diff
@@ -328,7 +334,13 @@ RUN export DEBIAN_FRONTEND=noninteractive \
         && if [ -d /usr/share/postgresql/9.5/extension ]; then \
             cd /usr/share/postgresql/9.5/extension \
             && for f in *.sql *.control; do \
-                for v in 9.3 9.4 9.6 10; do \
+                  if [ -L /usr/share/postgresql/9.5/extension/$f ]; then \
+                     continue; \
+                  fi; \
+                  for v in $PGOLDVERSIONS; do \
```
CyberDem0n: I would rather use
```
fi \
&& for v in $PGOLDVERSIONS; do \
```

postgres-appliance/Dockerfile.build
```diff
@@ -279,7 +279,13 @@ RUN export DEBIAN_FRONTEND=noninteractive \
         && if [ -d /usr/share/postgresql/9.5/contrib/postgis-$POSTGIS_VERSION ]; then \
             cd /usr/share/postgresql/9.5/contrib/postgis-$POSTGIS_VERSION \
             && for f in *.sql *.pl; do \
-                for v in 9.4 9.6 10; do \
+                 if [ -L /usr/share/postgresql/9.5/contrib/postgis-$POSTGIS_VERSION/$f ]; then \
+                   continue; \
+                 fi; \
```
CyberDem0n: I would rather use
```
fi \
&& for v in $PGOLDVERSIONS; do \
```

CyberDem0n:
:+1:

avaczi:
:+1:

CyberDem0n:
Merged, thank you!

PR #307, opened 2019-02-18
Fix check of AWS_SECRET_ACCESS_KEY and AWS_ACCESS_KEY_ID

CyberDem0n:
the structure should not only contain keys but values should be not empty.

In addition to that set the default value of WALG_S3_SSE to AES256 to have behavior similar to wal-e


Jan-M:
:+1:

PR #310, opened 2019-02-20
Get rid of patroni_aws

CyberDem0n:
It was used only on AWS for clusters without epi and setting wrong names on EBS volumes.


avaczi:
:+1:

avaczi:
:+1:

CyberDem0n:
:+1:

PR #320, opened 2019-03-06
use `wal-g delete retain` command to delete older backups

tnextday:
Both wal-g and wal-e support delete backups use `retain` mode.
references:
https://github.com/wal-g/wal-g#usage
https://github.com/wal-e/wal-e#delete

### DIFF
--- a/postgres-appliance/scripts/postgres_backup.sh
+++ b/postgres-appliance/scripts/postgres_backup.sh
@@ -38,25 +38,10 @@ else
     POOL_SIZE="--pool-size ${POOL_SIZE}"
 fi
 
-BEFORE=""
-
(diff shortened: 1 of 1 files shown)


CyberDem0n:
Hi @tnextday,

We want to retain backups for at least `$DAYS_TO_RETAIN` rather than keep a fixed amount of backups.

If the node was promoted to the master within [one hour](https://github.com/zalando/spilo/blob/master/postgres-appliance/launch.sh#L36), spilo will upload the new basebackup. Just think about a chain of failovers/switchovers happened on the same day, it could easily breach the retention policy.

tnextday:
It’s more appropriate to use time to delete it. I close the PR.

PR #325, opened 2019-03-23
Kubernetes Chart Improvement

k1ng440:
Hello,
This pull request contains improvements for the chart.

I have upgraded to latest versions of Kubernetes and much more readable.
also, I have added services for master and slave so the application can select master for write and slave for read

### DIFF
--- a/kubernetes/spilo_kubernetes.yaml
+++ b/kubernetes/spilo_kubernetes.yaml
@@ -1,4 +1,6 @@
-# Replace *cluster_name* with 
+# Replace *cluster_name* with your desired patroni cluster name
+# Please Note: This template is only an example. Please use this as referance
+
 
 apiVersion: apps/v1beta1
 kind: StatefulSet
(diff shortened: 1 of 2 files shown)


kubernetes/spilo_kubernetes.yaml
```diff
@@ -132,47 +127,77 @@ spec:
...
-    spilo-cluster: *cluster_name
+    spilo-cluster: *cluster_name*
 subsets: []
 
+# Services
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: *cluster_name*-master
```
CyberDem0n: The name of the master (leader) service MUST match with the cluster name. Label selector is redundant here, because Patroni controls the corresponding [Endpoint](https://github.com/zalando/spilo/pull/325/files#diff-1cc5f12e11c6450ce292458f7fb8426cR128) and the service works automatically.

kubernetes/spilo_kubernetes.yaml
```diff
@@ -132,47 +127,77 @@ spec:
...
+  selector:
+    application: spilo
+    spilo-cluster: *cluster_name*
+    role: master
+  ports:
+  - name: postgresql
+    port: 5432
+    targetPort: 5432
+  - name: patroni
+    port: 8008
```
CyberDem0n: We don't need and don't want to expose the Patroni REST API on the leader service.

kubernetes/spilo_kubernetes.yaml
```diff
@@ -132,47 +127,77 @@ spec:
...
+    targetPort: 8008
+
 ---
 apiVersion: v1
 kind: Service
 metadata:
-  name: &cluster_name zalandodemo01
+  name: *cluster_name*-slave
```
CyberDem0n: There are no slaves in the postgres world.

kubernetes/spilo_kubernetes.yaml
```diff
@@ -183,13 +208,13 @@ data:
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  name: operator
+  name: *cluster_name*
```
CyberDem0n: There is no need to create a separate ServiceAccount/Role/RoleBinding per postgres cluster.

kubernetes/spilo_kubernetes.yaml
k1ng440: what do you call read-only replica?

k1ng440:
@CyberDem0n please review the new chart.

PR #330, opened 2019-04-04
Enable extensions based on availability for the specific version

CyberDem0n:
* timescaledb is available only for 9.6, 10 and 11
* pg_cron starting from 9.5
* pg_partman and pg_stat_kcache starting from 9.4

such a "smart" configuration will work only if there are no `postgresql.parameters.shared_preload_libraries` or `postgresql.parameters.extwlist.extensions` defined in the `SPILO_CONFIGURATION`


avaczi:
:+1:

CyberDem0n:
:+1:

PR #336, opened 2019-04-29
Fix S3_HOST detection for non-AWS S3

IlyaSemenov:
When using non-AWS S3 provider, `S3_HOST` is constructed incorrectly in WAL restore script. Moreover, it is impossible to override `S3_HOST` with container environment, so there's no way for WAL restore to work with non-AWS S3 storage. See https://github.com/zalando/postgres-operator/issues/550 for details.

The PR fixes `S3_HOST` detection for non-AWS S3 endpoints. Of course, it still works for AWS S3.

### DIFF
--- a/postgres-appliance/scripts/wal-e-wal-fetch.sh
+++ b/postgres-appliance/scripts/wal-e-wal-fetch.sh
@@ -68,10 +68,13 @@ else
     exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


CyberDem0n:
:+1:

FxKu:
:+1:

IlyaSemenov:
Awesome! Thanks for the swift turnaround. I've noticed that this has already been published as registry.opensource.zalan.do/acid/spilo-cdp-11:1.5-p75, but I had to manually request [the API server](https://registry.opensource.zalan.do/v2/acid/spilo-cdp-11/tags/list) to discover that. Is there a page where new updates are published in human readable form?

Issue #350, opened 2019-07-19
master fails to build plprofiler


optix2000:
```
+ git clone -b REL3_5 https://github.com/pgcentral/plprofiler.git
Cloning into 'plprofiler'...
fatal: could not read Username for 'https://github.com': No such device or address
```
Looks like https://github.com/pgcentral/plprofiler.git no longer exists.

Issue #347, opened 2019-07-08
1.5-p9 Docker build fails again


alexzhc:
Hi, Zalando team, 

Building 1.5-p9 image from source code by default parameters fails at patching pgbouncer. Is it safe to skip the patching? Please help. 

Thanks
Alex

```
+ apt-get source pgbouncer
Reading package lists...
NOTICE: 'pgbouncer' packaging is maintained in the 'Git' version control system at:
https://salsa.debian.org/postgresql/pgbouncer.git
Please use:
git clone https://salsa.debian.org/postgresql/pgbouncer.git
to retrieve the latest (possibly unreleased) updates to the package.
Need to get 491 kB of source archives.
Get:1 http://apt.postgresql.org/pub/repos/apt bionic-pgdg/main pgbouncer 1.10.0-1.pgdg18.04+1 (dsc) [1373 B]
Get:2 http://apt.postgresql.org/pub/repos/apt bionic-pgdg/main pgbouncer 1.10.0-1.pgdg18.04+1 (tar) [481 kB]
Get:3 http://apt.postgresql.org/pub/repos/apt bionic-pgdg/main pgbouncer 1.10.0-1.pgdg18.04+1 (diff) [9548 B]
dpkg-source: info: extracting pgbouncer in pgbouncer-1.10.0
dpkg-source: info: unpacking pgbouncer_1.10.0.orig.tar.gz
dpkg-source: info: unpacking pgbouncer_1.10.0-1.pgdg18.04+1.debian.tar.xz
dpkg-source: info: applying debian-config
dpkg-source: info: applying python3
Fetched 491 kB in 2s (208 kB/s)
W: Download is performed unsandboxed as root as file 'pgbouncer_1.10.0-1.pgdg18.04+1.dsc' couldn't be accessed by user '_apt'. - pkgAcquire::Run (13: Permission denied)
+ ls -d pgbouncer-1.10.0
+ cd pgbouncer-1.10.0
+ curl -sL https://github.com/pgbouncer/pgbouncer/pull/127.diff
+ patch -p1
patching file include/bouncer.h
Hunk #1 FAILED at 230.
1 out of 1 hunk FAILED -- saving rejects to file include/bouncer.h.rej
patching file src/objects.c
Hunk #1 FAILED at 550.
Hunk #2 FAILED at 714.
Hunk #3 FAILED at 786.
Hunk #4 succeeded at 1121 (offset 58 lines).
Hunk #5 succeeded at 1190 (offset 58 lines).
[...]

PR #351, opened 2019-07-26
Fix image build

### LINKED ISSUE #283: Please add timescaledb to the list of shared_preload_libraries
Hi,
Although spilo comes with timescaledb built as part of the Docker image, timescaledb requires that the shared_preload_libraries parameter in postgresql.conf be updated with "timescaledb" extension, which is not being done currently. Updating postgresql.conf post-init is out of the question with Patroni managing the file.

Can we please add timescaledb to the list of extensions by default ? 
https://github.com/zalando/spilo/blob/master/postgres-appliance/scripts/configure_spilo.py#L209

Thanks.

CyberDem0n:
* pgbouncer package was upgraded and doesn't need one patch anymore
* plprofiler was moved to bigsql
* fix the build and the runtime with DEMO=true
* plpgsql_check 1.7.3
* timescaledb 1.4.0

Close #283, #347, #350

### DIFF
--- a/postgres-appliance/Dockerfile
+++ b/postgres-appliance/Dockerfile
@@ -1,8 +1,7 @@
 ARG PGVERSION=11
+ARG DEMO=false
 
 FROM ubuntu:18.04 as builder
-
-ARG DEMO=false
 
 RUN export DEBIAN_FRONTEND=noninteractive \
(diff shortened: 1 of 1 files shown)


avaczi:
:+1:

PR #354, opened 2019-08-12
Enable wal-g to be used by default when restoring from S3

CyberDem0n:
In addition to that switch to wal-g 0.2.11 and decrease shared_buffers on K8s


avaczi:
:+1:

CyberDem0n:
:+1:

PR #355, opened 2019-08-22
Set TIMESCALEDB env variable in the final docker image

CyberDem0n:
It is used by the post_init script in order to figure out if the extension needs to be upgraded.
If the env isn't set the script was trying to upgrade it on every failover/switchover even when it is not necessary.


avaczi:
:+1:

CyberDem0n:
:+1:

PR #364, opened 2019-09-25
Add IPv6 support to Spilo

optix2000:
Fixes zalando/postgres-operator#639

### DIFF
--- a/postgres-appliance/scripts/configure_spilo.py
+++ b/postgres-appliance/scripts/configure_spilo.py
@@ -389,6 +389,7 @@ def set_walg_placeholders(placeholders, prefix=''):
         value = str(placeholders.get(prefix + name, placeholders[prefix + 'USE_WALG'])).lower()
         placeholders[prefix + name] = 'true' if value == 'true' and walg_supported else None
 
+
 def get_listen_ip():
     """ Get IP to listen on for things that don't natively support detecting IPv4/IPv6 dualstack """
     def has_dual_stack():
@@ -399,7 +400,7 @@ def get_listen_ip():
                 sock.setsockopt(socket.IPPROTO_IPV6, socket.IPV6_V6ONLY, False)
                 return True
             except socket.error as e:
-                logger.debug('Error when working with ipv6 socket: %s', e)
+                logging.debug('Error when working with ipv6 socket: %s', e)
             finally:
                 if sock:
                     sock.close()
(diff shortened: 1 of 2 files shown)


postgres-appliance/scripts/configure_spilo.py
```diff
@@ -236,7 +236,7 @@ def deep_update(a, b):
     ssl_cert_file: {{SSL_CERTIFICATE_FILE}}
     ssl_key_file: {{SSL_PRIVATE_KEY_FILE}}
     shared_preload_libraries: 'bg_mon,pg_stat_statements,pgextwlist,pg_auth_mon,set_user'
-    bg_mon.listen_address: '0.0.0.0'
+    bg_mon.listen_address: '{{LISTEN_IP}}'
```
CyberDem0n: This one is the most interesting, because bg_mon doesn't handle it automatically yet, but we can handle it here.

Depending on situation the `LISTEN_IP` should be set either to `'::'` or to `'0.0.0.0'`.
There are 4 different cases:
1. IPv4 only => `'0.0.0.0'`
2. IPv6 only => `'::'`
3. Dual-stack is supported by OS => `'::'`. Listening on `'::'` will automatically make it accessible via IPv4
4. Dual-stack is not supported => `'0.0.0.0'`

The final function will look like:
```python
def bg_mon_listen_ip(port=8080):
    def has_dual_stack():
        if hasattr(socket, 'AF_INET6') and hasattr(socket, 'IPPROTO_IPV6') and hasattr(socket, 'IPV6_V6ONLY'):
            sock = None
            try:
                sock = socket.socket(socket.AF_INET6, socket.SOCK_STREAM)
                sock.setsockopt(socket.IPPROTO_IPV6, socket.IPV6_V6ONLY, False)
                return True
            except socket.error as e:
                logger.debug('Error when working with ipv6 socket: %s', e)
            finally:
                if sock:
                    sock.close()
        return False

    info = socket.getaddrinfo(None, port, socket.AF_UNSPEC, socket.SOCK_STREAM, 0, socket.AI_PASSIVE)
    # in case dual stack is not supported we want IPv4 to be preferred over IPv6
    info.sort(key=lambda x: x[0] == socket.AF_INET, reverse=not has_dual_stack())
    return info[0][-1][0]
```

Jan-M:
@CyberDem0n

Jan-M:
:+1:

CyberDem0n:
:+1:

PR #365, opened 2019-10-10
Fix postgis upgrade

FxKu:
Add symlink for old PostGIS version and update database extension if exists.

### DIFF
--- a/postgres-appliance/Dockerfile
+++ b/postgres-appliance/Dockerfile
@@ -212,6 +212,7 @@ RUN export DEBIAN_FRONTEND=noninteractive \
             if [ $version != "9.3" ] && [ $version != "9.4" ]; then \
                 apt-get install -y postgresql-${version}-postgis-${POSTGIS_VERSION%.0} \
                                 postgresql-${version}-postgis-${POSTGIS_VERSION%.0}-scripts; \
+                && ln -s /usr/lib/postgresql/$PGVERSION/lib/postgis-${POSTGIS_VERSION%.0}.so /usr/lib/postgresql/$PGVERSION/lib/postgis-2.5.so \
             fi \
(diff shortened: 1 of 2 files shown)


postgres-appliance/Dockerfile
```diff
@@ -212,6 +212,7 @@ RUN export DEBIAN_FRONTEND=noninteractive \
             if [ $version != "9.3" ] && [ $version != "9.4" ]; then \
                 apt-get install -y postgresql-${version}-postgis-${POSTGIS_VERSION%.0} \
                                 postgresql-${version}-postgis-${POSTGIS_VERSION%.0}-scripts; \
+                && ln -s /usr/lib/postgresql/$PGVERSION/lib/postgis-${POSTGIS_VERSION%.0}.so /usr/lib/postgresql/$PGVERSION/lib/postgis-2.5.so \
```
CyberDem0n: s/PGVERSION/version/g

postgres-appliance/scripts/post_init.sh
```diff
@@ -137,6 +137,11 @@ while IFS= read -r db_name; do
     if [ "x$TIMESCALEDB_VERSION" != "x" ] && [ "x$TIMESCALEDB_VERSION" != "x$TIMESCALEDB" ]; then
         echo "ALTER EXTENSION timescaledb UPDATE;"
     fi
+    POSTGISDB_VERSION=$(echo -e "SELECT NULL;\nSELECT extversion FROM pg_extension WHERE extname = 'postgis'" | psql -tAX -d ${db_name} 2> /dev/null | tail -n 1)
```
CyberDem0n: There are multiple issues.
1. `POSTGIS_VERSION` environment variable is not available here.
2. we probably need to run `ALTER EXTENSION` not only for `postgis`

PR #366, opened 2019-10-11
[configure_spilo] add ALLOW_NOSSL environment variable

### LINKED ISSUE #367: Add option to allow non-ssl connections
some apps (such as Geoserver) don't support connecting to the database over SSL. it would be really nice if we could disable this requirement with an environment variable

alice-sawatzky:
some apps (such as Geoserver) don't support connecting to the database over SSL. This adds an environment variable `ALLOW_NOSSL` in order to allow these applications to still connect to the database.

fixes #367

### DIFF
--- a/postgres-appliance/scripts/configure_spilo.py
+++ b/postgres-appliance/scripts/configure_spilo.py
@@ -251,15 +251,15 @@ postgresql:
     {{/PAM_OAUTH2}}
     - host    all             all                ::1/128            md5
     - hostssl replication     {{PGUSER_STANDBY}} all                md5
-    {{#ALLOW_NOSSL}}
-    - host    all             all                all                md5
-    {{/ALLOW_NOSSL}}
     {{^ALLOW_NOSSL}}
     - hostnossl all           all                all                reject
     {{/ALLOW_NOSSL}}
     {{#PAM_OAUTH2}}
     - hostssl all             +{{HUMAN_ROLE}}    all                pam
     {{/PAM_OAUTH2}}
+    {{#ALLOW_NOSSL}}
+    - host    all             all                all                md5
+    {{/ALLOW_NOSSL}}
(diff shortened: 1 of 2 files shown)


alice-sawatzky:
thanks, i missed that! i'll push a patch soon

alice-sawatzky:
should be fixed now, thanks for the feedback!

postgres-appliance/scripts/configure_spilo.py
```diff
@@ -251,7 +251,12 @@ def deep_update(a, b):
     {{/PAM_OAUTH2}}
     - host    all             all                ::1/128            md5
     - hostssl replication     {{PGUSER_STANDBY}} all                md5
+    {{#ALLOW_NOSSL}}
```
CyberDem0n: Unfortunately it is still wrong.

This block should be after(instead) of https://github.com/zalando/spilo/pull/366/files#diff-5e29afdf8025daafbc08cc254eb83f9fR263. The issue is that [PAM](https://github.com/zalando/spilo/pull/366/files#diff-5e29afdf8025daafbc08cc254eb83f9fR261) requires the password transmitted as it is, therefore connection must be encrypted.

The change https://github.com/zalando/spilo/pull/366/files#diff-5e29afdf8025daafbc08cc254eb83f9fR257-R259 is fine.

alice-sawatzky: ok, I've read the docs on pg_hba.conf and i think i understand what you're saying. sorry for all the back and forth on this!

alice-sawatzky:
awesome, thanks for the approve! any idea when this can be merged / make it into a release?

alice-sawatzky:
@CyberDem0n can you look into merging / moving forward the process for this to be merged?

erthalion:
:+1:

FxKu:
:+1:

PR #370, opened 2019-11-01
Switch from supervisord to runit

CyberDem0n:
It provides nearly the same functionality but the memory footprint is smaller.
In addition to that make Patroni process real-time and fix a little bug with IPv6 detection.


avaczi:
:+1:

CyberDem0n:
:+1:

PR #371, opened 2019-11-01
Run dumb-init only after decompressing image

CyberDem0n:
If the launch.sh is called with init parameter it will re-execute itself via dumb-init


sdudoladov:
:+1:

avaczi:
:+1:

PR #372, opened 2019-11-01
Patroni needs more environment variables

CyberDem0n:
For example KUBERNETES_SERVICE_* are necessary for communication with K8s API, POD_NAME and POD_NAMESPACE and needed for callback scripts, and so on.

We can't just pass all of them due to the https://github.com/zalando/patroni/commit/d770c910fdeb4420f4fd33c7f39f3f487a5cdf2b


sdudoladov:
:+1:

CyberDem0n:
:+1:

PR #373, opened 2019-11-03
handle TLS certificate rotation

zimbatm:
When the SSL_PRIVATE_KEY_FILE is not a self-generated certificate, the file can change over time. But right now postgresql doesn't know about it and won't load the new certificate automatically.

This PR introduces a bit of a polling approach to solve that problem. Every 5 minutes it will have a look if the certificate has changed within that interval and trigger a `pg_ctl reload` if yes. There are possibly smarter approaches like listen on inotify events but this has the advantage that no additional process needs to be running and running a bash script every 5 minutes is quite lite-weight.

### DIFF
--- a/postgres-appliance/scripts/configure_spilo.py
+++ b/postgres-appliance/scripts/configure_spilo.py
@@ -459,6 +459,7 @@ def get_placeholders(provider):
     placeholders.setdefault('SSL_CRL_FILE', '')
     placeholders.setdefault('SSL_CERTIFICATE_FILE', os.path.join(placeholders['PGHOME'], 'server.crt'))
     placeholders.setdefault('SSL_PRIVATE_KEY_FILE', os.path.join(placeholders['PGHOME'], 'server.key'))
+    placeholders.setdefault('SSL_TEST_RELOAD', os.environ.get('SSL_PRIVATE_KEY_FILE', '') != '')
     placeholders.setdefault('WALE_BACKUP_THRESHOLD_MEGABYTES', 102400)
     placeholders.setdefault('WALE_BACKUP_THRESHOLD_PERCENTAGE', 30)
     # if Kubernetes is defined as a DCS, derive the namespace from the POD_NAMESPACE, if not set explicitely.
@@ -785,6 +786,9 @@ def write_crontab(placeholders, overwrite):
     link_runit_service('cron')
 
     lines = ['PATH={PATH}'.format(**placeholders)]
+
+    if placeholders.get('SSL_TEST_RELOAD'):
+        lines += ['*/5 * * * * PGDATA={PGDATA} SSL_CA_FILE={SSL_CA_FILE} SSL_CRL_FILE={SSL_CRL_FILE} SSL_CERTIFICATE_FILE={SSL_CERTIFICATE_FILE} SSL_PRIVATE_KEY_FILE={SSL_PRIVATE_KEY_FILE} /scripts/test_reload_ssl.sh 5'.format(**placeholders)]
 
     if bool(placeholders.get('USE_WALE')):
         lines += [('{BACKUP_SCHEDULE} envdir "{WALE_ENV_DIR}" /scripts/postgres_backup.sh' +
--- a/postgres-appliance/scripts/configure_spilo.py
+++ b/postgres-appliance/scripts/configure_spilo.py
@@ -244,6 +244,7 @@ postgresql:
     log_truncate_on_rotation: 'on'
     ssl: 'on'
     ssl_ca_file: {{SSL_CA_FILE}}
+    ssl_crl_file: {{SSL_CRL_FILE}}
     ssl_cert_file: {{SSL_CERTIFICATE_FILE}}
     ssl_key_file: {{SSL_PRIVATE_KEY_FILE}}
     shared_preload_libraries: 'bg_mon,pg_stat_statements,pgextwlist,pg_auth_mon,set_user'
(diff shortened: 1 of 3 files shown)


CyberDem0n:
How the file is generated/updated? Maybe there is a way to run the script or call REST API endpoint after the update?
I am asking because the next Patroni release will always do `pg_ctl reload` when the `POST /reload` REST API call is received.

zimbatm:
In my case, the file is generated by the cert-manager operator on kubernetes. Cert-manager generates a Secret for each Certificate custom resource. Then the patroni container mounts that secret under `/tls` and has `SSL_PRIVATE_KEY_FILE` set to `/tls/tls.key` (among other things).

The nice thing about this design is that the interface is quite simple. The secret that patroni is mounting is of the same shape as if a user would have used `kubectl create secret tls ...`.
Cert-manager doesn't have to know about the consumers of the secret which keeps the coupling fairly low.

Cert-manager currently doesn't have a callback feature that could be used to trigger the `POST /reload` on certificate renewal.

zimbatm:
Thanks for the review. It definitely needs more testing before it gets merged but it seems like you agree with the overall design.

postgres-appliance/scripts/configure_spilo.py
```diff
@@ -780,7 +780,10 @@ def write_crontab(placeholders, overwrite):
 
     link_runit_service('cron')
 
-    lines = ['PATH={PATH}'.format(**placeholders)]
+    lines = [
+        'PATH={PATH}'.format(**placeholders),
+        '* * * * */5 PGDATA=${PGDATA} /scripts/test_reload.sh 5'.format(**placeholders),
```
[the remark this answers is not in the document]
zimbatm: Good catch. Fixed!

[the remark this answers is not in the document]
zimbatm: I also changed the logic a bit to only run the cron reload when the user passes a custom SSL_CERTIFICATE_FILE

postgres-appliance/scripts/test_reload.sh
```diff
@@ -0,0 +1,39 @@
...
+        return 2
+    fi
+    if [[ ! -e "$file" ]]; then
+        debug "$file does not exist"
+        return 2
+    fi
+    local mtime now
+    mtime=$(stat -c '%Y' "$file")
+    now=$(date +%s)
+    [[ $((now - 1 - mtime)) > $test_interval_minutes ]]
```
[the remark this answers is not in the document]
zimbatm: fixed!

zimbatm:
@CyberDem0n mind taking another look?

zimbatm:
I did some manual testing to check that the cron works properly. Is there a testing infrastructure that you use before creating a release?

zimbatm:
@CyberDem0n ping :)

postgres-appliance/scripts/test_reload_ssl.sh
```diff
@@ -0,0 +1,41 @@
...
+    local mtime now elapsed_sec
+    mtime=$(stat -c '%Y' "$file")
+    now=$(date +%s)
+    elapsed_sec=$(( now - 1 - mtime ))
+    [[ $elapsed_sec -lt $test_interval_sec ]]
+}
+
+## Main ##
+
+if has_changed_in_interval "$SSL_PRIVATE_KEY_FILE"; then
```
CyberDem0n: You [setup](https://github.com/zalando/spilo/pull/373/files#diff-5e29afdf8025daafbc08cc254eb83f9fR787) a cron job based on `SSL_CERTIFICATE_FILE`, but here you check the `SSL_PRIVATE_KEY_FILE`, which is rarely supposed to be rotated.

postgres-appliance/scripts/configure_spilo.py
```diff
@@ -782,6 +782,11 @@ def write_crontab(placeholders, overwrite):
 
     lines = ['PATH={PATH}'.format(**placeholders)]
 
+    # if SSL_CERTIFICATE_FILE is set then the ceritificate is provided by the
+    # user and might be rotated.
+    if os.environ.get('SSL_CERTIFICATE_FILE'):
+        lines += ['*/5 * * * * PGDATA={PGDATA} /scripts/test_reload_ssl.sh 5'.format(**placeholders)]
```
CyberDem0n: `PGDATA={PGDATA}` -- yeah, cron cleans the environment, we have to do the same for `SSL_*_FILE` and co.

zimbatm: done

postgres-appliance/scripts/test_reload_ssl.sh
zimbatm: doh. It now checks for an independent `SSL_TEST_RELOAD` attribute, which is set to True if `SSL_CERTIFICATE_FILE` is set or False otherwise to allow more flexibility and expose the logic.

zimbatm:
> What about merging #380 into #373 and also adding support for ssl_crl_file?
done
> These 4 files should be checked for update and reload performed.
done

I will do some testing tomorrow

zimbatm:
Tested manually:

* [ ] crontab is not added on empty environment
* [ ] crontab is added if SSL_PRIVATE_KEY_FILE or SSL_TEST_RELOAD is set
* [ ] test_reload_ssl.sh triggers if files mtime are younger than poll interval
* [ ] test_reload_ssl.sh doesn't trigger otherwise

CyberDem0n:
:+1:

FxKu:
:+1:

PR #375, opened 2019-11-06
Fix patroni runit script and add pgbouncer

CyberDem0n:
Patroni was failing to bootstrap due to the wrong HOME directory


avaczi:
:+1:

CyberDem0n:
:+1:

PR #378, opened 2019-11-12
Fix clone with pg_upgrade

CyberDem0n:
it requires PGHOME env and future Patroni requires modul import in a specific order


avaczi:
:+1:

CyberDem0n:
:+1:

PR #380, opened 2019-11-14
add SSL_CA_FILE configuration

zimbatm:
allow to set the CA if you want to handle SSL client validation


zimbatm:
merging with #373

PR #395, opened 2019-12-12
Solve problem with backup_label

CyberDem0n:
remove the file if we can proof that postgres was really in backup mode


sdudoladov:
:+1:

CyberDem0n:
:+1:

PR #397, opened 2019-12-18
Bump the default backup retention to 5

CyberDem0n:
keeping backups for the past 5 days gives us a guaranty that on Monday we can recover from some failure on Thursday.


FxKu:
:+1:

CyberDem0n:
:+1:

PR #400, opened 2019-12-23
Timescaledb 1.5.1

CyberDem0n:
Timescaledb is planning to deprecate support for postgres 9.6 in upcoming releases. In order to prepare for it we introduce the new variable, `TIMESCALEDB_LEGACY`. Now it has the same value as
`TIMESCALEDB`, but in future it will stick to the last release supporting 9.6.

In addition to that fix the postgis upgrade and reorder PATH. It is better when the psql is found as /usr/bin/psql, because it does some magic with LD_PRELOAD=ibreadline.so in order to provide better usability.


FxKu:
:+1:

CyberDem0n:
:+1:

PR #401, opened 2019-12-23
Limit the number of unexpected restarts of Patroni

CyberDem0n:
Usually that happens if the custom_bootstrap is failing.

The maximum number of restarts is 5. Between every attempt the backoff is applied. Backoff time is calculated as ATTEMPT*30.

Cases when Patroni is killed with -9 or by OOM are excluded (number of restarts is not limited and backoff is not applied).


avaczi:
:+1:

CyberDem0n:
:+1:

PR #404, opened 2020-01-23
Added missing reference when using basebackup_fast_xlog method.

ermajn:
Template have `basebackup_fast_xlog` defined but there was missing reference to a command.


ermajn:
> The `bootstrap.dcs.standby_cluster.create_replica_methods.basebackup_fast_xlog` refers already existing section `postgresql.basebackup_fast_xlog`. It is located just a few lines up from this place.

You are refering to the one that is under USE_WALE? I want to cover the case when I don't want to use WAL-E

ermajn:
in a line https://github.com/zalando/spilo/blob/09b4cc3c18c0b1d8e666157171794a2a181cdc14/postgres-appliance/scripts/configure_spilo.py#L144
basebackup_fast_xlog is defined regardless of using STANDBY_WITH_WALE. 
So in a case when standby is defined with no WALE, basebackup_fast_xlog have no reference to a script. 
Perhaps my ference should be under STANDBY_CLUSTER section not STANDBY_HOST. I will fix that. Please let me know if this is ok, as my patch works nice in my scenario.

postgres-appliance/scripts/configure_spilo.py
```diff
@@ -314,6 +314,13 @@ def deep_update(a, b):
     command: /scripts/basebackup.sh
     retries: 2
 {{/USE_WALE}}
+{{#STANDBY_CLUSTER}}
+  create_replica_method:
```
CyberDem0n: You are right, if `USE_WALE` is not set, there will be no `postgresql.basebackup_fast_xlog`.

But, we don't want to have some duplicates in the config if both, `USE_WALE` and `STANDBY_CLUSTER` are set.

ermajn:
@CyberDem0n I think `basebackup_fast_xlog:` should be always referenced with latest commit.

CyberDem0n:
@ermajn sorry, I don't get what you are trying to say.
also currently PR is a mess, 16 changed files.

ermajn:
Its a merge of my fork with master. I can sreate a separate, as changes in my PR is only in one file.

Issue #374, opened 2019-11-04
Spilo with non-root User and Read Only Root Volume


guruguruguru:
Hey,

i wanted to ask if you see a good possibility to run Spilo with non-root User and Root Read Only.
Our Kubernetes Cluster does not allow root or writeable root volumes. What I can do is using EmptyDir and Persistent Volumes.

Any idea on this topic?

Thanks,
Dennis

flickerfly:
This is important for running on Red Hat OpenShift as well. You can modify things to work around it, but it breaks best practices.

One place that is mentioned outside just OpenShift is: https://learnk8s.io/production-best-practices/#governance

CyberDem0n:
Agree, it would be great, but it will require quite a few changes.
Basically all config files (Patroni, wal-e/wal-g, cron, and so on) should be stored on `/run`.
Recently I changed `supervisord` to `runit` and it already works with `/run/service` instead of `/etc/service`.
It is not really our priority and somebody should just continue working on it.

flickerfly:
@CyberDem0n thanks for moving the ball along. I appreciate knowing you are aware of it and watching for opportunities to improve towards that as you handle the higher priorities. That makes me comfortable enough to work with workarounds and look forward to the improvements.

ReSearchITEng:
For running in OpenShift, no root and no special permissions required.
Operator Image should be at least:  registry.opensource.zalan.do/acid/postgres-operator:v1.4.0-21-g1249626-dirty
Operator should be configured with these values:
  kubernetes_use_configmaps: "true"
  docker_image: registry.opensource.zalan.do/acid/spilo-cdp-12:1.6-p114 #or newer

To get the latest versions of the images of both operator and spilo, do:
https://registry.opensource.zalan.do/v2/acid/postgres-operator/tags/list
https://registry.opensource.zalan.do/v2/acid/spilo-cdp-12/tags/list

PR #405, opened 2020-01-29
Make it possible to run with read-only root fs

CyberDem0n:
It includes moving all writable files into the `/run` directory.
In case if the file still must be located in the usual place we create a symlink.
When starting with read-only option in addition to the `/home/postgres/pgdata` volume one has to provide writable (tmpfs) `/run` volume.

Example:
```
$ docker run -v $(pwd)/pgdata:/home/postgres/pgdata \
    --tmpfs /run:rw,noexec,nosuid --read-only spilo-12
```
In addition to that, improve the `postgres_backup.sh` script. It will receive `BACKUP_NUM_TO_RETAIN` from the envdir instead of parameter.

Fixes https://github.com/zalando/spilo/issues/374


avaczi:
:+1:

erthalion:
:+1:

CyberDem0n:
:+1:

Issue #407, opened 2020-02-12
Cut a new release


birdayz:
It would be great to have a release containing& supporting Patroni 1.6.4. Thanks!

CyberDem0n:
We are working on it, but it makes sense to wait till tomorrow, because of postgres minor releases.

marcostvz:
Hi @CyberDem0n, today we tried to upgrade to spilo-12, but it failed due to the bug on multi-lined SPILO_CONFIGURATION env var fixed in https://github.com/zalando/spilo/commit/46fbf78eaef7100188c3dc1a675a27c0b3a63bb5

Are there plans for a new spilo-12 release? (no pressure, I just want to know about timing to organize myself)

Kind regards :)

FxKu:
Included in [1.6-p3](https://github.com/zalando/spilo/releases/tag/1.6-p3)

PR #408, opened 2020-02-12
Prepare next release

### LINKED ISSUE #406: Using WALE_S3_ENDPOINT causes backup failures in newer S3 regions
The wal-e library interprets the presence of WALE_S3_ENDPOINT to mean a non-Amazon S3 implementation:
>     # Insist that one of AWS_REGION or WALE_S3_ENDPOINT is defined.
>     # The former is for authenticating correctly with AWS SigV4.
>     #
>     # The latter is for alternate implementations that are
>     # S3-interface compatible.  Many, or most, of these do not support
>     # AWS SigV4 at all and none are known to require SigV4 (and
>     # instead use the non-region-demanding SigV2), so simplify by
>     # relaxing the AWS_REGION requirement in that case.
(https://github.com/wal-e/wal-e/blob/2bf9df9cd82f1600495f2cbb6621ee4c18d92cc6/wal_e/blobstore/s3/s3_util.py#L171)

This causes failures in newer AWS regions (those > 2013) as those regions insist on the signature v4 (sigv4) implementation, which is usable in every AWS region unlike the older sigv2. This will be come a bigger problem later this year:

> Revised Plan – Any new buckets created after June 24, 2020 will not support SigV2 signed requests, although existing buckets will continue to support SigV2 while we work with customers to move off this older request signing method.
(https://aws.amazon.com/blogs/aws/amazon-s3-update-sigv2-deprecation-period-extended-modified/)

I've worked around by commenting out lines 714-715 in configure_spilo.py https://github.com/zalando/spilo/blob/master/postgres-appliance/scripts/configure_spilo.py#L714

CyberDem0n:
* Build the image without compression by default and keep all layers when compression is not used.
* Bump versions of Patroni, Debezium, bg_mon, Timescaledb, and plpgsql_check.
* Set `AWS_REGION` only for AWS S3. For S3-compatible storages set `WALE_S3_ENDPOINT` and `AWS_ENDPOINT`.
* Avoid producing ERROR when trying to upgrade unexisting PostGIS.
*  Keep one backup more. For big clusters the backup jobs take a significant time what causing the recovery window to become smaller than a threshold.
* Temporary configure `vm.dirty_*bytes`.

Close #406, #407

### DIFF
--- a/postgres-appliance/Dockerfile
+++ b/postgres-appliance/Dockerfile
@@ -1,9 +1,10 @@
 ARG PGVERSION=12
-ARG TIMESCALEDB=1.5.1
+ARG TIMESCALEDB=1.6.0
 ARG TIMESCALEDB_LEGACY=$TIMESCALEDB
 ARG DEMO=false
+ARG COMPRESS=false
 
-FROM ubuntu:18.04 as builder
+FROM ubuntu:18.04 as builder-false
 
 RUN export DEBIAN_FRONTEND=noninteractive \
     && echo 'APT::Install-Recommends "0";\nAPT::Install-Suggests "0";' > /etc/apt/apt.conf.d/01norecommend \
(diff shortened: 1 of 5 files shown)


Jan-M:
:+1:

CyberDem0n:
:+1:

Issue #384, opened 2019-11-21
New spilo from the most recent master crashes


gmldba:
I just cloned the repo and build the image . When I use it in the postgres-operator, it crashes. Am I missing something?

CyberDem0n:
Please define "crashes" and provide some evidence, i.e. logs

gmldba:
From docker
`docker run -it ************/spilo:1.6-p1-13-gb75771e
root@a9f7c5777c92:/# psql
Warning: No existing local cluster is suitable as a default target. Please see man pg_wrapper(1) how to specify one.
Error: You must install at least one postgresql-client-<version> package
root@a9f7c5777c92:/# which psql
/usr/bin/psql
root@a9f7c5777c92:/# cat /var/log/
alternatives.log  bootstrap.log     dpkg.log          lastlog           tallylog
apt/              btmp              faillog           postgresql/       wtmp
root@a9f7c5777c92:/# cat /var/log/apt/
eipp.log.xz  history.log  term.log
root@a9f7c5777c92:/# cat /var/log/apt/history.log
root@a9f7c5777c92:/# cat /var/log/lastlog
root@a9f7c5777c92:/# cat /var/log/faillog`
From Kubernetes (postgres-operator)
`Name:           platops-test-cluster-0
Namespace:      default
Priority:       0
Node:           aks-agentpool-30264578-1/10.240.0.6
Start Time:     Tue, 26 Nov 2019 18:21:53 +0000
Labels:         application=spilo
                controller-revision-hash=platops-test-cluster-849cf9f68d
                statefulset.kubernetes.io/pod-name=platops-test-cluster-0
                team=PLATOPS
                version=platops-test-cluster
Annotations:    <none>
Status:         Running
IP:             10.244.0.229
IPs:            <none>
Controlled By:  StatefulSet/platops-test-cluster
Init Containers:
  date:
    Container ID:  docker://a3a3a61df751f4a3fdb8686a9be189dd2ea228d29e5036ada70d4fed0e1b8b6f
    Image:         busybox
    Image ID:      docker-pullable://busybox@sha256:1303dbf110c57f3edf68d9f5a16c082ec06c4cf7604831669faf2c712260b5a0
    Port:          <none>
    Host Port:     <none>
    Command:
      /bin/date
    State:          Terminated
      Reason:       Completed
      Exit Code:    0
      Started:      Tue, 26 Nov 2019 18:22:32 +0000
[...]

PR #412, opened 2020-02-18
Download Plantuner from HTTPS URL and pin version by commit hash

gertvdijk:
Before this change:
- Any version on default HEAD would be taken and thus a non-reproducible
  build.
- Insecure download over old-style Git protocol.

I was unable to find a stable HTTPS-powered mirror of the upstream
repository git://www.sigaev.ru/plantuner.git, so I created my own on
GitHub.
I did consider https://github.com/postgrespro/plantuner, but found its
state debatable as expressed in an issue filed there:
https://github.com/postgrespro/plantuner/issues/1#issue-566917641.

Fixes zalando/spilo#384


postgres-appliance/Dockerfile
```diff
@@ -96,7 +97,9 @@ RUN export DEBIAN_FRONTEND=noninteractive \
         && git clone -b $DECODERBUFS https://github.com/debezium/postgres-decoderbufs.git \
         && git clone -b $PLPROFILER https://github.com/bigsql/plprofiler.git \
         && git clone -b $PLPGSQL_CHECK https://github.com/okbob/plpgsql_check.git \
-        && git clone git://www.sigaev.ru/plantuner.git \
+        && mkdir plantuner \
+            && curl -sL https://github.com/gertvdijk/plantuner/archive/$PLANTUNER_COMMIT.tar.gz \
```
CyberDem0n: I would better prefer keeping the tarball in the `postgres-appliance/dependencies/`

gertvdijk: Okay, I will change that. :) Thank for your review.

avaczi:
:+1:

CyberDem0n:
:+1:

PR #414, opened 2020-02-24
allow rootless user

ToroNZ:
This PR allows dynamically assigned UID, the appropriate permissions required to run.

When not run under root (UID=0):
- "$RW_DIR" gets mounted under "$PGHOME/rw_dir"
- "$RW_DIR" and "$PGHOME" are owned by GID 0
- "/usr/sbin/cron" can be run by group (to allow any user access to cron)
- "/etc/passwd" is updated to reflect assigned UID for user postgres
- "write_certificates" function in 'configure_spilo.py' obtains postgres UID and sets key ownership
- SSL CERT/KEY moves to "$RW_DIR/certs" (could move back to the root of $RW_DIR but this way is easier/tidy (?) to overwrite)

I've tested this in Kubernetes, Docker and Openshift using UID  0, 101 and 1000320000. Both 'COMPRESS=TRUE' (root) and 'COMPRESS=FALSE' work fine. I haven't truly tested 
'DEMO=true', could do so later today.

### DIFF
--- a/postgres-appliance/Dockerfile
+++ b/postgres-appliance/Dockerfile
@@ -379,7 +379,7 @@ RUN export DEBIAN_FRONTEND=noninteractive \
 # Install patroni, wal-e and wal-g
 ENV PATRONIVERSION=1.6.3
 ENV WALE_VERSION=1.1.0
-ENV WALG_VERSION=v0.2.13
+ENV WALG_VERSION=v0.2.14
 RUN export DEBIAN_FRONTEND=noninteractive \
     && set -ex \
     && BUILD_PACKAGES="python3-pip python3-wheel python3-dev git patchutils binutils" \
--- a/postgres-appliance/Dockerfile
+++ b/postgres-appliance/Dockerfile
@@ -1,8 +1,10 @@
-ARG PGVERSION=11
-ARG TIMESCALEDB=1.4.2
+ARG PGVERSION=12
+ARG TIMESCALEDB=1.6.0
+ARG TIMESCALEDB_LEGACY=$TIMESCALEDB
 ARG DEMO=false
+ARG COMPRESS=false
 
-FROM ubuntu:18.04 as builder
+FROM ubuntu:18.04 as builder-false
 
 RUN export DEBIAN_FRONTEND=noninteractive \
     && echo 'APT::Install-Recommends "0";\nAPT::Install-Suggests "0";' > /etc/apt/apt.conf.d/01norecommend \
@@ -34,12 +36,6 @@ RUN export DEBIAN_FRONTEND=noninteractive \
     done \
     && curl -s -o - https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add - \
 \
-    && apt-get update \
-    && apt-get install -y postgresql-common \
-\
-    # forbid creation of a main cluster when package is installed
-    && sed -ri 's/#(create_main_cluster) .*$/\1 = false/' /etc/postgresql-common/createcluster.conf \
-\
     # Clean up
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/* \
@@ -51,23 +47,25 @@ RUN export DEBIAN_FRONTEND=noninteractive \
     && find /var/log -type f -exec truncate --size 0 {} \;
 
 COPY dependencies/debs /builddeps
-COPY set_user.patch /builddeps
+ADD set_user.patch /builddeps/
 
 ARG PGVERSION
 ARG TIMESCALEDB
+ARG TIMESCALEDB_LEGACY
 ARG DEMO
-ARG PGOLDVERSIONS="9.3 9.4 9.5 9.6 10"
+ARG COMPRESS
+ARG PGOLDVERSIONS="9.3 9.4 9.5 9.6 10 11"
 ARG WITH_PERL=false
 
 ARG DEB_PG_SUPPORTED_VERSIONS="$PGOLDVERSIONS $PGVERSION"
 
 # Install PostgreSQL, extensions and contribs
-ENV POSTGIS_VERSION=2.5 \
-    BG_MON_COMMIT=0df2a6f7ebcaf805465271032d1ec788779938d7 \
+ENV POSTGIS_VERSION=3.0 \
+    BG_MON_COMMIT=7a4a4a621f89cc032bd7d497ce2373a1220c89a5 \
     PG_AUTH_MON_COMMIT=902c670c9f2071b30e1d4437e7a38e2e136f9be6 \
-    DECODERBUFS=v0.10.0.Final \
+    DECODERBUFS=v1.0.1.Final \
     SET_USER=REL1_6_2 \
-    PLPGSQL_CHECK=v1.7.6 \
+    PLPGSQL_CHECK=v1.8.2 \
     PLPROFILER=REL4_1 \
     PAM_OAUTH2=v1.0
 
@@ -77,30 +75,18 @@ RUN export DEBIAN_FRONTEND=noninteractive \
     && apt-get update \
     && cd /builddeps \
 \
-    && BUILD_PACKAGES="devscripts build-essential fakeroot debhelper git gcc libc6-dev make cmake libevent-dev naturaldocs python libssl-dev libkrb5-dev" \
-# pgq3: naturaldocs python
+    && BUILD_PACKAGES="devscripts equivs build-essential fakeroot debhelper git gcc libc6-dev make cmake libevent-dev libssl-dev libkrb5-dev" \
     && if [ "$DEMO" = "true" ]; then \
         export DEB_PG_SUPPORTED_VERSIONS="$PGVERSION" \
+        && WITH_PERL=false \
+        && rm -f *.deb \
         && apt-get install -y $BUILD_PACKAGES; \
     else \
-        BUILD_PACKAGES="$BUILD_PACKAGES equivs pgxnclient zlib1g-dev libprotobuf-c-dev liblwgeom-dev libproj-dev libxslt1-dev libxml2-dev libpam0g-dev libedit-dev libselinux1-dev libcurl4-openssl-dev libicu-dev libc-ares-dev pandoc pkg-config" \
+        BUILD_PACKAGES="$BUILD_PACKAGES pgxnclient zlib1g-dev libprotobuf-c-dev liblwgeom-dev libproj-dev libxslt1-dev libxml2-dev libpam0g-dev libedit-dev libselinux1-dev libcurl4-openssl-dev libicu-dev python libc-ares-dev pandoc pkg-config" \
 # debezium-decoderbufs: libprotobuf-c-dev liblwgeom-dev libproj-dev
 # pgbouncer: libc-ares-dev pandoc pkg-config
 # pg_rewind: libxslt1-dev libxml2-dev libpam0g-dev libkrb5-dev libedit-dev libselinux1-dev
         && apt-get install -y $BUILD_PACKAGES libprotobuf-c1 libcurl4 \
-\
-        && if [ "$WITH_PERL" != "true" ]; then \
-            version=$(apt-cache show perl | sed -n 's/^Version: //p' | sort -rV | head -n 1) \
-            && echo "Section: misc\nPriority: optional\nStandards-Version: 3.9.8\nPackage: perl\nSection:perl\nMulti-Arch: allowed\nReplaces: perl-base\nVersion: $version\nDescription: perl" > perl \
-            && equivs-build perl; \
-        fi \
-\
-        && for p in python3-keyring python3-docutils ieee-data; do \
-            version=$(apt-cache show $p | sed -n 's/^Version: //p' | sort -rV | head -n 1) \
-            && echo "Section: misc\nPriority: optional\nStandards-Version: 3.9.8\nPackage: $p\nVersion: $version\nDescription: $p" > $p \
-            && equivs-build $p; \
-        done \
-        && dpkg -i *.deb || apt-get -y -f install \
 \
         # install pam_oauth2.so
         && git clone -b $PAM_OAUTH2 --recurse-submodules https://github.com/CyberDem0n/pam-oauth2.git \
@@ -110,16 +96,34 @@ RUN export DEBIAN_FRONTEND=noninteractive \
         && git clone -b $DECODERBUFS https://github.com/debezium/postgres-decoderbufs.git \
         && git clone -b $PLPROFILER https://github.com/bigsql/plprofiler.git \
         && git clone -b $PLPGSQL_CHECK https://github.com/okbob/plpgsql_check.git \
-        && git clone git://www.sigaev.ru/plantuner.git; \
+        && git clone git://www.sigaev.ru/plantuner.git \
+\
+        && for p in python3-keyring python3-docutils ieee-data; do \
+            version=$(apt-cache show $p | sed -n 's/^Version: //p' | sort -rV | head -n 1) \
+            && echo "Section: misc\nPriority: optional\nStandards-Version: 3.9.8\nPackage: $p\nVersion: $version\nDescription: $p" > $p \
+            && equivs-build $p; \
+        done; \
     fi \
+
+\
+    && if [ "$WITH_PERL" != "true" ]; then \
+        version=$(apt-cache show perl | sed -n 's/^Version: //p' | sort -rV | head -n 1) \
+        && echo "Section: misc\nPriority: optional\nStandards-Version: 3.9.8\nPackage: perl\nSection:perl\nMulti-Arch: allowed\nReplaces: perl-base\nVersion: $version\nDescription: perl" > perl \
+        && equivs-build perl; \
+    fi \
+\
+    && if [ "$WITH_PERL" != "true" ] || [ "$DEMO" != "true" ]; then dpkg -i *.deb || apt-get -y -f install; fi \
 \
     && curl -sL https://github.com/CyberDem0n/bg_mon/archive/$BG_MON_COMMIT.tar.gz | tar xz \
     && curl -sL https://github.com/RafiaSabih/pg_auth_mon/archive/$PG_AUTH_MON_COMMIT.tar.gz | tar xz \
     && git clone -b $SET_USER https://github.com/pgaudit/set_user.git \
     && patch -d set_user -p1 < set_user.patch \
-    && git clone -b $TIMESCALEDB https://github.com/timescale/timescaledb.git \
+    && git clone -b $TIMESCALEDB_LEGACY https://github.com/timescale/timescaledb.git \
 \
-    && apt-get install -y libevent-2.1 libevent-pthreads-2.1 python3.6 python3-psycopg2 \
+    && apt-get install -y postgresql-common libevent-2.1 libevent-pthreads-2.1 python3.6 python3-psycopg2 \
+\
+    # forbid creation of a main cluster when package is installed
+    && sed -ri 's/#(create_main_cluster) .*$/\1 = false/' /etc/postgresql-common/createcluster.conf \
 \
     && for version in $DEB_PG_SUPPORTED_VERSIONS; do \
             sed -i "s/ main.*$/ main $version/g" /etc/apt/sources.list.d/pgdg.list \
@@ -136,12 +140,13 @@ RUN export DEBIAN_FRONTEND=noninteractive \
                 && if [ "$WITH_PERL" = "true" ]; then \
                     EXTRAS="$EXTRAS postgresql-plperl-${version}"; \
                 fi \
-                && if [ $version != "9.3" ]; then \
-                    EXTRAS="$EXTRAS \
-                            postgresql-${version}-pglogical \
-                            postgresql-${version}-postgis-${POSTGIS_VERSION} \
-                            postgresql-${version}-postgis-${POSTGIS_VERSION}-scripts \
-                            postgresql-${version}-wal2json" \
+                && if [ $version != "9.3" ]; then EXTRAS="$EXTRAS postgresql-${version}-wal2json"; fi \
+                && if [ $version != "9.3" ] && [ $version != "9.4" ]; then \
+                    EXTRAS="$EXTRAS postgresql-${version}-postgis-${POSTGIS_VERSION%.0} \
+                            postgresql-${version}-postgis-${POSTGIS_VERSION%.0}-scripts"; \
+                fi \
+                && if [ $version != "9.3" ] && [ $version != "12" ]; then \
+                    EXTRAS="$EXTRAS postgresql-${version}-pglogical" \
                     && if [ $version != "9.4" ]; then \
                         EXTRAS="$EXTRAS postgresql-${version}-pgl-ddl-deploy"; \
                     fi \
@@ -165,15 +170,16 @@ RUN export DEBIAN_FRONTEND=noninteractive \
             # Install 3rd party stuff
             && if [ "$version" = "9.6" ] || [ "$version" = "10" ] || [ "$version" = "11" ] ; then \
                 cd timescaledb \
+                && if [ "$version" != "9.6" ]; then git checkout $TIMESCALEDB; fi \
                 && rm -fr build \
-                && ./bootstrap -DAPACHE_ONLY=1 -DSEND_TELEMETRY_DEFAULT=NO \
+                && ./bootstrap -DAPACHE_ONLY=1 -DSEND_TELEMETRY_DEFAULT=NO -DREGRESS_CHECKS=OFF \
                 && make -C build install \
                 && strip /usr/lib/postgresql/$version/lib/timescaledb*.so \
                 && cd ..; \
             fi \
 \
             && if [ "$DEMO" != "true" ]; then \
-                if [ "$version" != "11" ]; then \
+                if [ "$version" != "11" ] && [ $version != 12 ]; then \
                     for extension in quantile trimmed_aggregates; do \
                         pgxn install $extension \
                         && strip /usr/lib/postgresql/$version/lib/$extension.so; \
@@ -217,7 +223,12 @@ RUN export DEBIAN_FRONTEND=noninteractive \
         for version in $DEB_PG_SUPPORTED_VERSIONS; do \
             if [ $version != "9.3" ]; then \
                 # due to dependency issues partman has to be installed separately
-                apt-get install -y postgresql-${version}-partman; \
+                apt-get install -y postgresql-${version}-partman \
+                && if [ $version != "9.4" ]; then \
+                    # create postgis symlinks to make it possible to perform update
+                    ln -s postgis-${POSTGIS_VERSION%.0}.so \
+                        /usr/lib/postgresql/${version}/lib/postgis-2.5.so; \
+                fi; \
             fi; \
         done \
         # patch, build and install pgbouncer
@@ -251,7 +262,7 @@ RUN export DEBIAN_FRONTEND=noninteractive \
     done \
 \
     # Remove unnecessary packages
-    && apt-get purge -y ${BUILD_PACKAGES} postgresql postgresql-server-dev-* libmagic1 bsdmainutils \
+    && apt-get purge -y ${BUILD_PACKAGES} postgresql postgresql-server-dev-* libpq-dev=$PGVERSION* libmagic1 bsdmainutils \
     && apt-get autoremove -y \
     && apt-get clean \
     && dpkg -l | grep '^rc' | awk '{print $2}' | xargs apt-get purge -y \
@@ -268,6 +279,7 @@ RUN export DEBIAN_FRONTEND=noninteractive \
             done; \
         done \
 \
+        && PGVERSION=11 \
         && cd /usr/share/postgresql/$PGVERSION/contrib/postgis-$POSTGIS_VERSION \
         && for f in *.sql *.pl; do \
             for v in /usr/share/postgresql/*; do \
@@ -377,7 +389,7 @@ RUN export DEBIAN_FRONTEND=noninteractive \
     && find /var/log -type f -exec truncate --size 0 {} \;
 
 # Install patroni, wal-e and wal-g
-ENV PATRONIVERSION=1.6.3
+ENV PATRONIVERSION=1.6.4
 ENV WALE_VERSION=1.1.0
 ENV WALG_VERSION=v0.2.14
 RUN export DEBIAN_FRONTEND=noninteractive \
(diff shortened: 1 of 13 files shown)


postgres-appliance/Dockerfile
```diff
@@ -513,9 +515,19 @@ WORKDIR $PGHOME
 COPY motd /etc/
 COPY runit /etc/service/
 COPY pgq_ticker.ini $PGHOME/
+COPY scripts bootstrap /scripts/
```
CyberDem0n: Why this one was moved from https://github.com/zalando/spilo/pull/414/files#diff-82f985fcf1262fb2d3d1843b9e74d9c3L542?

postgres-appliance/Dockerfile
```diff
@@ -513,9 +515,19 @@ WORKDIR $PGHOME
...
 
 RUN sed -i "s|/var/lib/postgresql.*|$PGHOME:/bin/bash|" /etc/passwd \
-        && chown -R postgres:postgres $PGHOME $RW_DIR \
+        && if [ "$COMPRESS" != "true" ]; then \
+           # non-root prep
+           mkdir -p $PGHOME \
+           && sed -i "s|/var/lib/postgresql.*|$PGHOME:/bin/bash|" /etc/passwd \
+           # Set permissions
+           && chmod 775 $PGHOME \
```
CyberDem0n: PGHOME already exists after https://github.com/zalando/spilo/pull/414/files#diff-82f985fcf1262fb2d3d1843b9e74d9c3R517 and its permissions already 755

The `/etc/passwd` already contains the correct home directory for postgres: https://github.com/zalando/spilo/pull/414/files#diff-82f985fcf1262fb2d3d1843b9e74d9c3R520

postgres-appliance/Dockerfile
```diff
@@ -513,9 +515,19 @@ WORKDIR $PGHOME
...
+           && chmod 664 /etc/passwd \
+           && mkdir -p $PGHOME/rw_dir \
+           && ln -s $PGHOME/rw_dir $RW_DIR/postgresql; \
```
CyberDem0n: I still don't get the purpose of `$PGHOME/rw_dir`.

postgres-appliance/launch.sh
```diff
@@ -1,5 +1,12 @@
 #!/bin/sh
 
+if [ "$(id -u)" -ne 0 ]; then
+    sed -e "s/^postgres:x:[^:]*:[^:]*:/postgres:x:$(id -u):$(id -g):/" /etc/passwd > /tmp/passwd
+    cat /tmp/passwd > /etc/passwd
+    rm /tmp/passwd
+    mkdir -p "$PGROOT" "$RW_DIR/cron/crontabs"
```
CyberDem0n: Isn't [configure_spilo.py](https://github.com/zalando/spilo/blob/master/postgres-appliance/scripts/configure_spilo.py#L790-L794) supposed to do it?

postgres-appliance/launch.sh
```diff
@@ -26,13 +33,19 @@ for i in $(seq 0 7); do
...
-elif python3 /scripts/configure_spilo.py all; then
-    su postgres -c "PATH=$PATH /scripts/patroni_wait.sh -t 3600 -- envdir $WALE_ENV_DIR /scripts/postgres_backup.sh $PGDATA $BACKUP_NUM_TO_RETAIN" &
+else
+    if [ "$(id -u)" -ne 0 ]; then
+        python3 /scripts/configure_spilo.py all
+        PATH=$PATH /scripts/patroni_wait.sh -t 3600 -- envdir $WALE_ENV_DIR /scripts/postgres_backup.sh $PGDATA $BACKUP_NUM_TO_RETAIN &
+    else
+        python3 /scripts/configure_spilo.py all
+        su postgres -c "PATH=$PATH /scripts/patroni_wait.sh -t 3600 -- envdir $WALE_ENV_DIR /scripts/postgres_backup.sh $PGDATA $BACKUP_NUM_TO_RETAIN" &
+    fi
```
CyberDem0n: Here you are changing the logic.
The original code is checking exit code of configure_spilo.py

postgres-appliance/scripts/configure_spilo.py
```diff
@@ -106,7 +107,7 @@ def write_certificates(environment, overwrite):
         output, _ = p.communicate()
         logging.debug(output)
 
-    uid = os.stat(environment['PGHOME']).st_uid
+    uid = pwd.getpwnam(environment['PGUSER_SUPERUSER']).pw_uid
```
CyberDem0n: This one seems to be totally unnecessary, the original code is producing the same result.
```
$ docker run -u 1001 --rm -ti --entrypoint /bin/bash registry.opensource.zalan.do/acid/spilo-12:1.6-p2 
I have no name!@48fb2b152229:/home/postgres$ python3
Python 3.6.9 (default, Nov  7 2019, 10:44:02) 
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.stat('/home/postgres').st_uid
101
>>>
```

postgres-appliance/launch.sh
CyberDem0n: The ` $BACKUP_NUM_TO_RETAIN` argument could be dropped (I totally forgot about it).

postgres-appliance/Dockerfile
ToroNZ: My bad, is a leftover from one of my tests where I was declaring USER at the bottom. I'll revert this back.

postgres-appliance/Dockerfile
ToroNZ: You're right, I'll remove this (just tested it okay).

postgres-appliance/Dockerfile
ToroNZ: I thought it was better to get it out of /run. I'll get rid of it, it is purely cosmetic anyway.

postgres-appliance/launch.sh
ToroNZ: Will fix the 'elif' logic and drop `$BACKUP_NUM_TO_RETAIN`

postgres-appliance/scripts/configure_spilo.py
ToroNZ: The reason behind this was to obtain the UID from the system not the filesystem.
Once `/etc/passwd` is set, results are different:

```
$ docker run -ti --rm --user 1000320000 --entrypoint /bin/bash toronz/spilo-min-cdp-12:1.6-v1.3.9.9
I have no name!@2ababdef33fa:/home/postgres$ python3
Python 3.6.9 (default, Nov  7 2019, 10:44:02) 
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.stat('/home/postgres').st_uid
101
>>> import pwd
>>> pwd.getpwnam('postgres').pw_uid
101
>>> exit()
I have no name!@2ababdef33fa:/home/postgres$ if [ "$(id -u)" -ne 0 ]; then
>     sed -e "s/^postgres:x:[^:]*:[^:]*:/postgres:x:$(id -u):$(id -g):/" /etc/passwd > /tmp/passwd
>     cat /tmp/passwd > /etc/passwd
>     rm /tmp/passwd
>     mkdir -p "$PGROOT"
> fi
I have no name!@2ababdef33fa:/home/postgres$ python3
Python 3.6.9 (default, Nov  7 2019, 10:44:02) 
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.stat('/home/postgres').st_uid
101
>>> import pwd
>>> pwd.getpwnam('postgres').pw_uid
1000320000
```

I used my image based on this PR, due to the permissions on /etc/passwd.

ToroNZ:
Went through it and amended based on the comments. Keen to hear thoughts regarding the permissions for the SSL key.

postgres-appliance/launch.sh
```diff
@@ -26,13 +33,15 @@ for i in $(seq 0 7); do
         touch "${PGLOG}/postgresql-$i.csv"
     fi
 done
-chown -R postgres:postgres "$PGROOT" "$RW_DIR/postgresql"
+chown -R postgres: "$PGROOT" "$RW_DIR/certs"
```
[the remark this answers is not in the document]
ToroNZ: Because it `$RW_DIR` already has the appropriate permissions based on group membership and `$RW_DIR/postgresql` is created by the run user here https://github.com/ToroNZ/spilo/blob/fd579c00b7895c8b12a8df746a667e7921b9d319/postgres-appliance/launch.sh#L27

postgres-appliance/launch.sh
```diff
@@ -26,13 +33,15 @@ for i in $(seq 0 7); do
...
 chmod 01777 "$RW_DIR/tmp"
 
 if [ "$DEMO" = "true" ]; then
     python3 /scripts/configure_spilo.py patroni pgqd certificate pam-oauth2
+elif [ "$(id -u)" -ne 0 ] && python3 /scripts/configure_spilo.py all; then
+    PATH=$PATH /scripts/patroni_wait.sh -t 3600 -- envdir $WALE_ENV_DIR /scripts/postgres_backup.sh $PGDATA &
 elif python3 /scripts/configure_spilo.py all; then
-    su postgres -c "PATH=$PATH /scripts/patroni_wait.sh -t 3600 -- envdir $WALE_ENV_DIR /scripts/postgres_backup.sh $PGDATA $BACKUP_NUM_TO_RETAIN" &
+    su postgres -c "PATH=$PATH /scripts/patroni_wait.sh -t 3600 -- envdir $WALE_ENV_DIR /scripts/postgres_backup.sh $PGDATA" &
```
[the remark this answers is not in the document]
ToroNZ: sweet as

postgres-appliance/Dockerfile
```diff
@@ -537,9 +543,18 @@ RUN sed -i "s|/var/lib/postgresql.*|$PGHOME:/bin/bash|" /etc/passwd \
...
-        && echo "source /etc/motd" >> /root/.bashrc
+        && echo "source /etc/motd" >> /root/.bashrc \
+        # Allow users in the root group to access the following dirs
+        && if [ "$COMPRESS" != "true" ]; then \
+           chgrp -R 0 $RW_DIR \
+           && chmod -R g=u $RW_DIR \
+           && chgrp -R 0 $PGHOME \
+           && chmod -R g=u $PGHOME \
+           && usermod -a -G root postgres \
+           && chmod gu+s /usr/sbin/cron; \
```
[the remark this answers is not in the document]
ToroNZ: I see what you mean, but the problem of user 'postgres' having had UID 101 assigned will not change based on group membership.

We could drop `import pwd` and use:
```
uid = os.stat(os.path.join(environment['RW_DIR'], 'certs')).st_uid
os.chmod(environment['SSL_PRIVATE_KEY_FILE'], 0o600)
os.chown(environment['SSL_PRIVATE_KEY_FILE'], uid, -1)
```
The above works (thanks to this [line](https://github.com/ToroNZ/spilo/blob/6ca00c8f4cd58a6f776955355bbda3b3b5071ffe/postgres-appliance/launch.sh#L27) and [line](https://github.com/ToroNZ/spilo/blob/6ca00c8f4cd58a6f776955355bbda3b3b5071ffe/postgres-appliance/launch.sh#L36)).

Let me know what you think.

Samusername:
--> ToronNZ:master --> I downloaded the whole ToronNZ:master.
I changed entrypoint to USER 101 in the end of Dockerfile, just before last CMD line.
( I do not know is there some other configuration needed for non-root execution in spilo. )

- docker build, docker run starts, on laptop. 
(A "seteuid" notification is not shown (anymore) with that Dockerfile/docker image ).
Tested with docker 1.13.1, and with docker 18.09.7. Tested also with --drop-cap=all. "seteuid" notification is not shown. (Good.)
- In a k8s lab (which uses docker 1.13.1), 
I see following in kubectl log output: "seteuid: Operation not permitted". I started the spilo / db cluster with pg-operator, spiloFSGroup: 103. 
(Same result without spiloFSGroup).

ToroNZ:
@Samusername - Thanks for testing. You should be able to test this PR by just adding "--user 101" to the `docker run` command.

The only Kubernetes I could find with such an old Docker version (v1.13) was an Openshift v3.11 cluster, which seems to be working just fine (this is assigning a dynamic UID out of 1008050000/10000). Which also means it runs under the following conditions:

- Capabilities: none
- Privileged: False
- SELINUX: MustRunAs
- RUNASUSER: MustRunAsRange
- FSGROUP: MustRunAs
- SUPGROUP: RunAsAny
- READONLYROOTFS: False

I just built a new image, updated the statefulset image tag, deleted a replica pod and here's the output of the new one coming up (Openshift v3.11 as above):

```
2020-03-02 21:40:25,494 - bootstrapping - INFO - Figuring out my environment (Google? AWS? Openstack? Local?)
2020-03-02 21:40:27,501 - bootstrapping - INFO - Could not connect to 169.254.169.254, assuming local Docker setup
2020-03-02 21:40:27,502 - bootstrapping - INFO - No meta-data available for this provider
2020-03-02 21:40:27,503 - bootstrapping - INFO - Looks like your running local
2020-03-02 21:40:27,514 - bootstrapping - WARNING - could not parse kubernetes labels as a JSON: Expecting value: line 1 column 1 (char 0), reverting to the default: {"application": "spilo"}
2020-03-02 21:40:27,529 - bootstrapping - INFO - Configuring pgqd
2020-03-02 21:40:27,530 - bootstrapping - INFO - Configuring certificate
2020-03-02 21:40:27,530 - bootstrapping - INFO - Generating ssl certificate
2020-03-02 21:40:27,693 - bootstrapping - INFO - Configuring patroni
2020-03-02 21:40:27,700 - bootstrapping - INFO - Writing to file /home/postgres/postgres.yml
2020-03-02 21:40:27,714 - bootstrapping - INFO - Configuring pam-oauth2
2020-03-02 21:40:27,715 - bootstrapping - INFO - Writing to file /etc/pam.d/postgresql
2020-03-02 21:40:27,715 - bootstrapping - INFO - Configuring pgbouncer
2020-03-02 21:40:27,715 - bootstrapping - INFO - No PGBOUNCER_CONFIGURATION was specified, skipping
2020-03-02 21:40:27,715 - bootstrapping - INFO - Configuring bootstrap
2020-03-02 21:40:27,715 - bootstrapping - INFO - Configuring crontab
2020-03-02 21:40:27,715 - bootstrapping - INFO - Configuring log
2020-03-02 21:40:27,715 - bootstrapping - INFO - Configuring wal-e
2020-03-02 21:40:27,715 - bootstrapping - INFO - Configuring renice
2020-03-02 21:40:27,718 - bootstrapping - INFO - Skipping creation of renice cron job due to lack of permissions
2020-03-02 21:40:27,719 - bootstrapping - INFO - Configuring standby-cluster
2020-03-02 21:40:29,418 INFO: No PostgreSQL configuration items changed, nothing to reload.
2020-03-02 21:40:29,437 WARNING: Postgresql is not running.
2020-03-02 21:40:29,438 INFO: Lock owner: acid-tomas-demo9000-0; I am acid-tomas-demo9000-1
2020-03-02 21:40:29,443 INFO: pg_controldata:
  pg_control version number: 1201
  Catalog version number: 201909212
  Database system identifier: 6797547306099626073
  Database cluster state: shut down in recovery
  pg_control last modified: Mon Mar  2 21:40:04 2020
  Latest checkpoint location: 0/8000060
  Latest checkpoint's REDO location: 0/7001FF8
  Latest checkpoint's REDO WAL file: 000000030000000000000007
  Latest checkpoint's TimeLineID: 3
  Latest checkpoint's PrevTimeLineID: 3
  Latest checkpoint's full_page_writes: on
  Latest checkpoint's NextXID: 0:1193
  Latest checkpoint's NextOID: 40960
  Latest checkpoint's NextMultiXactId: 1
  Latest checkpoint's NextMultiOffset: 0
  Latest checkpoint's oldestXID: 480
  Latest checkpoint's oldestXID's DB: 1
  Latest checkpoint's oldestActiveXID: 1193
  Latest checkpoint's oldestMultiXid: 1
  Latest checkpoint's oldestMulti's DB: 1
  Latest checkpoint's oldestCommitTsXid: 0
  Latest checkpoint's newestCommitTsXid: 0
  Time of latest checkpoint: Sun Mar  1 09:17:43 2020
  Fake LSN counter for unlogged rels: 0/3E8
  Minimum recovery ending location: 0/9000000
  Min recovery ending loc's timeline: 3
  Backup start location: 0/0
  Backup end location: 0/0
  End-of-backup record required: no
  wal_level setting: replica
  wal_log_hints setting: on
  max_connections setting: 100
  max_worker_processes setting: 8
  max_wal_senders setting: 10
  max_prepared_xacts setting: 0
  max_locks_per_xact setting: 64
  track_commit_timestamp setting: off
  Maximum data alignment: 8
  Database block size: 8192
  Blocks per segment of large relation: 131072
  WAL block size: 8192
  Bytes per WAL segment: 16777216
  Maximum length of identifiers: 64
  Maximum columns in an index: 32
  Maximum size of a TOAST chunk: 1996
  Size of a large-object chunk: 2048
  Date/time type storage: 64-bit integers
  Float4 argument passing: by value
  Float8 argument passing: by value
  Data page checksum version: 1
  Mock authentication nonce: a6a3c5b15a7425161624a4f986924880fea164deb2648860d320dbc798ad39fd

2020-03-02 21:40:29,444 INFO: Lock owner: acid-tomas-demo9000-0; I am acid-tomas-demo9000-1
2020-03-02 21:40:29,480 INFO: Local timeline=3 lsn=0/9000000
2020-03-02 21:40:29,495 INFO: master_timeline=3
2020-03-02 21:40:29,499 INFO: master: history=1	0/5000000	no recovery target specified

2	0/6000000	no recovery target specified

2020-03-02 21:40:29,499 INFO: Lock owner: acid-tomas-demo9000-0; I am acid-tomas-demo9000-1
2020-03-02 21:40:29,569 INFO: starting as a secondary
2020-03-02 21:40:30,169 INFO: postmaster pid=73
2020-03-02 21:40:30 UTC [73]: [1-1] 5e5d7d4e.49 0     LOG:  Auto detecting pg_stat_kcache.linux_hz parameter...
2020-03-02 21:40:30 UTC [73]: [2-1] 5e5d7d4e.49 0     LOG:  pg_stat_kcache.linux_hz is set to 1000000
2020-03-02 21:40:30 UTC [73]: [3-1] 5e5d7d4e.49 0     LOG:  starting PostgreSQL 12.2 (Ubuntu 12.2-1.pgdg18.04+1) on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 7.4.0-1ubuntu1~18.04.1) 7.4.0, 64-bit
2020-03-02 21:40:30 UTC [73]: [4-1] 5e5d7d4e.49 0     LOG:  listening on IPv4 address "0.0.0.0", port 5432
2020-03-02 21:40:30 UTC [73]: [5-1] 5e5d7d4e.49 0     LOG:  listening on IPv6 address "::", port 5432
/var/run/postgresql:5432 - no response
2020-03-02 21:40:30 UTC [73]: [6-1] 5e5d7d4e.49 0     LOG:  listening on Unix socket "/var/run/postgresql/.s.PGSQL.5432"
2020-03-02 21:40:30 UTC [73]: [7-1] 5e5d7d4e.49 0     LOG:  redirecting log output to logging collector process
2020-03-02 21:40:30 UTC [73]: [8-1] 5e5d7d4e.49 0     HINT:  Future log output will appear in directory "../pg_log".
/var/run/postgresql:5432 - accepting connections
/var/run/postgresql:5432 - accepting connections
2020-03-02 21:40:31,339 INFO: Lock owner: acid-tomas-demo9000-0; I am acid-tomas-demo9000-1
2020-03-02 21:40:31,339 INFO: does not have lock
2020-03-02 21:40:31,339 INFO: establishing a new patroni connection to the postgres cluster
  File "/scripts/callback_endpoint.py", line 9, in <module>
    from kubernetes import client as k8s_client, config as k8s_config
ModuleNotFoundError: No module named 'kubernetes'
2020-03-02 21:40:31,493 INFO: no action.  i am a secondary and i am following a leader
2020-03-02 21:40:41,339 INFO: Lock owner: acid-tomas-demo9000-0; I am acid-tomas-demo9000-1
```
Running user:
```
$ whoami
postgres
$ id -u
1000570000
```
Pod status:
```
NAME                    READY     STATUS    RESTARTS   AGE
acid-tomas-demo9000-0   1/1       Running   0          1m
acid-tomas-demo9000-1   1/1       Running   0          3m
acid-tomas-demo9000-2   1/1       Running   0          2m
```

Perhaps there's something else in your setup? What kind of storage are you using?

Samusername:
@ToroNZ Thanks for showing that it works in Openshift.
The "seteuid" notification works with (an older) version (of k8s setup). The "seteuid" problem started to occur with a more latter one. I will try to investigate it somehow here.

Samusername:
@ToroNZ Thanks for the test regarding Openshift!
Perhaps that Openshift system still allows something more than our system? I am not able to explain the difference otherwise.

I started "grepping" patroni source codes. First match for setuid or seteuid is in
extras/startup-scripts/patroni.upstart.conf
...
setuid postgres
setgid postgres
...
Can that explain continuous notifications regarding "seteuid: Operation not permitted". 
Could there be other places which trigger seteuid (or setuid) calls. (I will start and continue to browse / grep.)

Samusername:
With spilo 1.6-p2, I need to use following settings:

```
        securityContext:
          privileged: true
          capabilities:
            add:
              - ALL # ALL did not help with official spilo1.6-p2.
          runAsNonRoot: false
          runAsUser: 0 # 101 did not work.
          fsGroup: 0 # Was 103. (May not matter in this context.)
```

With the "ToroNZ:master" version (from day or two ago)
docker run works nicely ("seteuid" notifications are not shown there).

Also in k8s, the current "ToroNZ:master" fixes help some, like have already proven (e.g. in Openshift).
 
Following settings already work in my k8s system also (not Openshift).
(seteuid notification is not shown.)

```
        securityContext:
          privileged: true
          capabilities:
            drop:
              - ALL # This may not affect.
          runAsNonRoot: true
          runAsUser: 101
          fsGroup: 103
```
Already these kinds of improvements are be good ( ! ).

Samusername:
Tried to test ToroNZ:master (git clone today).
With image from "./build.sh ... COMPRESS=false", the process seems to start, 
although there are some serious looking notifications in kubectl logs <pod>:
```
mkdir: cannot create directory ‘/home/postgres/pgdata/pgroot’: Permission denied
mkdir: cannot create directory ‘/home/postgres/pgdata/pgroot’: Permission denied
mkdir: cannot create directory ‘/home/postgres/pgdata/pgroot’: Permission denied
touch: cannot touch '/home/postgres/pgdata/pgroot/pg_log/postgresql-0.csv': No such file or directory
touch: cannot touch '/home/postgres/pgdata/pgroot/pg_log/postgresql-1.csv': No such file or directory
touch: cannot touch '/home/postgres/pgdata/pgroot/pg_log/postgresql-2.csv': No such file or directory
touch: cannot touch '/home/postgres/pgdata/pgroot/pg_log/postgresql-3.csv': No such file or directory
touch: cannot touch '/home/postgres/pgdata/pgroot/pg_log/postgresql-4.csv': No such file or directory
touch: cannot touch '/home/postgres/pgdata/pgroot/pg_log/postgresql-5.csv': No such file or directory
touch: cannot touch '/home/postgres/pgdata/pgroot/pg_log/postgresql-6.csv': No such file or directory
touch: cannot touch '/home/postgres/pgdata/pgroot/pg_log/postgresql-7.csv': No such file or directory
chown: cannot access '/home/postgres/pgdata/pgroot': No such file or directory
2020-03-04 13:55:39,933 - bootstrapping - INFO - Figuring out my environment (Google? AWS? Openstack? Local?)
2020-03-04 13:55:41,940 - bootstrapping - INFO - Could not connect to <ipaddress>, assuming local Docker setup
2020-03-04 13:55:41,941 - bootstrapping - INFO - No meta-data available for this provider
2020-03-04 13:55:41,942 - bootstrapping - INFO - Looks like your running local
2020-03-04 13:55:41,973 - bootstrapping - INFO - Configuring renice
2020-03-04 13:55:41,977 - bootstrapping - INFO - Skipping creation of renice cron job due to lack of permissions
2020-03-04 13:55:41,978 - bootstrapping - INFO - Configuring pgbouncer
2020-03-04 13:55:41,978 - bootstrapping - INFO - No PGBOUNCER_CONFIGURATION was specified, skipping
2020-03-04 13:55:41,978 - bootstrapping - INFO - Configuring pgqd
2020-03-04 13:55:41,978 - bootstrapping - INFO - Configuring crontab
2020-03-04 13:55:41,978 - bootstrapping - INFO - Configuring log
2020-03-04 13:55:41,978 - bootstrapping - INFO - Configuring certificate
2020-03-04 13:55:41,978 - bootstrapping - INFO - Generating ssl certificate
2020-03-04 13:55:42,157 - bootstrapping - INFO - Configuring pam-oauth2
2020-03-04 13:55:42,157 - bootstrapping - INFO - No PAM_OAUTH2 configuration was specified, skipping
2020-03-04 13:55:42,157 - bootstrapping - INFO - Configuring standby-cluster
2020-03-04 13:55:42,157 - bootstrapping - INFO - Configuring wal-e
2020-03-04 13:55:42,157 - bootstrapping - INFO - Configuring patroni
2020-03-04 13:55:42,165 - bootstrapping - INFO - Writing to file /home/postgres/postgres.yml
2020-03-04 13:55:42,166 - bootstrapping - INFO - Configuring bootstrap
2020-03-04 13:55:43,573 INFO: No PostgreSQL configuration items changed, nothing to reload.
2020-03-04 13:55:43,576 INFO: Lock owner: None; I am patronidbcluster-0
2020-03-04 13:55:43,683 INFO: waiting for leader to bootstrap
2020-03-04 13:55:53,577 INFO: Lock owner: None; I am patronidbcluster-0
2020-03-04 13:55:53,577 INFO: waiting for leader to bootstrap
2020-03-04 13:56:03,577 INFO: Lock owner: None; I am patronidbcluster-0
...
```
I do not know could there be something wrong how I build the image (with help of Dockerfile). --> I am new to all of this / to the project.
I will try with pure "docker build" also.

I used securityContext, which was mentioned in my previous message.

When I try with COMPRESS=true, then following error is shown:
```
/launch.sh: line 8: can't create /etc/passwd: Permission denied
mkdir: can't create directory '/home/postgres/pgdata/pgroot': Permission denied
decompressing spilo image...
failed to decompress spilo image
```

ToroNZ:
@Samusername, are you using block storage? Any chance you can test this with file storage (i.e.: NFS) ? I think that's where your problem is. 
Try setting it up with `emptyDir{}` volumes if you don't have NFS or any other ReadWriteMany (https://kubernetes.io/docs/concepts/storage/volumes/#emptydir).

`COMPRESS=true` will only work if you're running as privileged user (root). This is a clever hack to archive the rootfs which makes the image lighter to move around.... though you loose boot-up speed on runtime due to the decompress stage.

postgres-appliance/Dockerfile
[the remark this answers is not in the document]
CyberDem0n: 467-468 seems to be useless.

Samusername:
In our k8s system, the db cluster has worked well, with non-root user 
(postges, 101, user), (spiloFSGroup: 103).
Testing only: inserts to single or two tables, delete pod randomly/all pods at once. Leader-sync(strict)-async. For couple of days.

spilo_privileged: "true", in postgres-operator. (In our k8s system the privileged setting seems to be needed: because of the seteuid notification.)
ToroNZ:master from last week.

This would already be significantly better (than earlier), also in our k8s system, it seems.

ReSearchITEng:
We've build the image using your PR on an openshift 4.2 without privileged mode.
We've added also the KUBERNETES_USE_CONFIGMAPS: "true"
All permissions issues seem to be gone.

Still, something is not working: the cluster goes in error: CreateFailed.
Operator complains it's not able to connect to the service to create users/etc.
There are no errors in the pod logs, and connecting to the pod I can run pql and connect locally.

(btw, should anyone else want to test the image based on this PR as of 18 March, is here:
docker pull docker.io/researchiteng/spilo:latest )

postgres-appliance/Dockerfile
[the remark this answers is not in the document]
ReSearchITEng: I have checked and these were solved and I tested and looks ok.
Can you mark this as solved?

Samusername:
The permission problems seems similar to: https://github.com/zalando/postgres-operator/issues/821
There is workaround, which helps. (Although it is not a solution.)
I think it happened with some other release also.

Samusername:
I have merged (only) the proposed changes from https://github.com/zalando/spilo/pull/414 to spilo master a little over week ago, etc. It worked ok, as nonRoot user. That is already a big improvement, and could be available for official use in master (if not there yet?). 
For my k8s system, small improvement would still be needed, I needed to use: spilo_privileged mode, like I wrote.  (I have not had time to experience with my own modifications regarding these topics.)  But such addiotional step could be done separately later also.

ReSearchITEng:
@Samusername 
> The permission problems seems similar to: [zalando/postgres-operator#821](https://github.com/zalando/postgres-operator/issues/821)
> There is workaround, which helps. (Although it is not a solution.)
> I think it happened with some other release also.
Nice trick with the init containers. Still, IMO it's reasonable to consider updating launch.sh as suggested in PR: https://github.com/ToroNZ/spilo/pull/1
My only question remains, between these options, (of which tested a&b and I confirm work fine), what would be the approved one, so I can adapt the PR accordingly: 
a. `chmod go-w "$PGDATA" || true`      # This is very "cautious" but only solves the error, no security fix inside it
b. `chmod -R go-w "$PGDATA" || true`  # This both solves the error and secures the files inside
c. `chmod -R go-w "$PGROOT" || true` # This solves the error and secures everything, skipping what cannot be changed. -> but might be risky. 
I would recommend 2nd one, shall I push the PR to the master directly?

ToroNZ:
> @Samusername
> regarding the merge to master spilo: I did not find the required OpernShift (dynamic userID) updates in master's launch.sh or Dockerfile. All the key parts are missing. It might be it's ok on K8s nonRoot, but OpenShift non-root/non-priviledged works only only if this PR is merged.
> 
> Regarding the init container workaround [[zalando/postgres-operator#821](https://github.com/zalando/postgres-operator/issues/821)]
> Nice trick. Still, IMO it's reasonable to consider updating launch.sh as suggested in PR: [ToroNZ#1](https://github.com/ToroNZ/spilo/pull/1)
> My only question remains, between these options, (of which tested a&b and I confirm work fine), what would be the approved one, so I can adapt the PR accordingly:
> a. `chmod go-w "$PGDATA" || true` # This is very "cautious" but only solves the error, no security fix inside it
> b. `chmod -R go-w "$PGDATA" || true` # This both solves the error and secures the files inside
> c. `chmod -R go-w "$PGROOT" || true` # This solves the error and secures everything, skipping what cannot be changed. -> but might be risky.
> I would recommend 2nd one, shall I push the PR to the master directly?

I've now included that in this PR :)

Thank you!

Jan-M:
Thank you for all the work put into this PR.

Jan-M:
:+1:

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you for your effort @ToroNZ, @Samusername, @ReSearchITEng

ReSearchITEng:
@ToroNZ @Samusername 
For running in OpenShift, no root and no special permissions required.
Operator Image should be at least:  registry.opensource.zalan.do/acid/postgres-operator:v1.4.0-21-g1249626-dirty
Operator should be configured with these values:
  kubernetes_use_configmaps: "true"
  docker_image: registry.opensource.zalan.do/acid/spilo-cdp-12:1.6-p114 #or newer

To get the latest versions of the images of both operator and spilo, do:
https://registry.opensource.zalan.do/v2/acid/postgres-operator/tags/list
https://registry.opensource.zalan.do/v2/acid/spilo-cdp-12/tags/list

PR #418, opened 2020-03-09
Renice without root

CyberDem0n:
1. Setup the renice cron job as a postgres user if the container is running with CAP_SYS_NICE. In order to be able to set negative and rt priorities without root, we set cap_sys_nice on chrt and renice tools.

2. The SUID bit on cron daemon makes it possible to start it from non-root user, what will help to simplify #414

3. Unrelated change: added tablefunc to the extwlist.extensions


FxKu:
:+1:

CyberDem0n:
:+1:

PR #419, opened 2020-03-12
add TIMESCALEDB_APACHE_ONLY docker build flag

db-adrian:
The TIMESCALEDB_APACHE_ONLY flag enables building the community timescaledb version.


postgres-appliance/Dockerfile
```diff
@@ -1,6 +1,7 @@
 ARG PGVERSION=12
 ARG TIMESCALEDB=1.6.0
 ARG TIMESCALEDB_LEGACY=$TIMESCALEDB
+ARG TIMESCALEDB_APACHE_ONLY=true
```
CyberDem0n: This line is not necessary.

postgres-appliance/Dockerfile
```diff
@@ -52,6 +53,7 @@ ADD set_user.patch /builddeps/
 ARG PGVERSION
 ARG TIMESCALEDB
 ARG TIMESCALEDB_LEGACY
+ARG TIMESCALEDB_APACHE_ONLY
```
CyberDem0n: TIMESCALEDB_APACHE_ONLY=true

postgres-appliance/Dockerfile
```diff
@@ -488,6 +490,7 @@ LABEL maintainer="Alexander Kukushkin <alexander.kukushkin@zalando.de>"
...
```
CyberDem0n: this line is not necessary.

avaczi:
:+1:

CyberDem0n:
:+1:

haf:
Do you host a version of this where the flag is set to false instead?

Because the difference between the apache 2 license and the community license only is that they don't want AWS/Azure/GCP to use it, how about defaulting this flag to "false", since the vast majority of the users here will never own a cloud of their own?

haf:
Trying to build the Dockerfile in git tag 1.6-p5 with above-mentioned license flag causes this error:

```
+ apt-get install --allow-downgrades -y postgresql-contrib-12 postgresql-plpython3-12 libpq5=12* postgresql-pltcl-12 postgresql-12-cron postgresql-12-hypopg postgresql-12-pgaudit postgresql-12-pg-checksums postgresql-12-pgl-ddl-deploy postgresql-12-pglogical postgresql-12-pglogical-ticker postgresql-12-pgq3 postgresql-12-pg-stat-kcache postgresql-12-pldebugger postgresql-12-pllua postgresql-12-plproxy postgresql-12-postgis-3 postgresql-12-postgis-3-scripts postgresql-12-repack postgresql-12-wal2json libpq-dev=12* postgresql-server-dev-12
Reading package lists...
Building dependency tree...
Reading state information...
E: Version '12*' for 'libpq5' was not found
E: Version '12*' for 'libpq-dev' was not found
```

CyberDem0n:
@haf https://github.com/zalando/spilo/issues?q=is%3Aissue+is%3Aclosed

haf:
@CyberDem0n How about this suggestion? https://github.com/zalando/spilo/pull/419#issuecomment-705616209 — making the default false?

CyberDem0n:
They want to protect not just from aws/gcp/azure, but any database as a service.
In some way, if you run postgres-operator + spilo you get exactly it, database as a service, so personally I a bit afraid of doing that.

A stable release is a hard topic. There are a lot of deb packages without pinned versions. It makes it easier to build an image for the next Postgres minor release, but at the same time increasing the risk of future build failures. Pinning to specific versions drastically increasing maintenance costs.

haf:
I wouldn't say that a normal failover-aware setup means I offer, to a market of buyers, database as a service for commercial gain. But if I host Spilo and TimescaleDB myself and sell that offering as is, without building an app around it, I'd be violating it.

How about writing a build manifest of all installed packages, post-install, and after release making that possible to install from, as part of the git tag? With e.g. `apt-clone`: https://unix.stackexchange.com/questions/191662/how-do-i-replicate-installed-package-selections-from-one-debian-system-to-anothe/191676#191676 — the list of packages could then be written into the command to install them.

I realise this is a burden to implement; there are simply no lockfiles on OS:es' packages right now. :-s

Here's an image from master: `haaf/spilo:1.6-timescale-cl-p5-master` on docker hub for anyone venturing here later. No guarantees on this image though, I may delete it, or Docker Inc may.

PR #421, opened 2020-03-20
Handle the case when value has newlines

CyberDem0n:
* `printenv -0` outputs env with null byte as delimiter
* `tr` removes all newlines
* `sed` replaces null bytes with newlines


sdudoladov:
:+1:

CyberDem0n:
:+1:

PR #423, opened 2020-03-26
enable Dockerfile build without build.sh

ReSearchITEng:
Trying to build spilo from docker hub fails at this last COPY as scm-source.json is dynamically created only when run by build.sh.
By adding "*" at the end of the file name, COPY will not fail and spilo can have image build by docker hub on docker.io automatically (at least for tests). This would increase the userbase of spilo image, as people are able to share the exact tested image without asking others to go though rebuild process just to test their PR.


ReSearchITEng:
@CyberDem0n - can you please help review this?

CyberDem0n:
I know that you don't want to touch the current workflow, but in fact there is no need to do so.
Our CI/CD system already automatically generates and injects scm-source.json into the image during the docker push.
The only reason having the build.sh script was to simplify the local build and push to our docker registry, and we are not pushing local builds anymore due to some compliance issues.

postgres-appliance/Dockerfile
```diff
@@ -564,6 +564,6 @@ RUN sed -i "s|/var/lib/postgresql.*|$PGHOME:/bin/bash|" /etc/passwd \
         fi
 
 COPY scripts bootstrap /scripts/
-COPY scm-source.json launch.sh /
+COPY scm-source.json* launch.sh /
```
CyberDem0n, line 567:
```suggestion
COPY launch.sh /
```

ReSearchITEng: I thought we want to keep this as an optional, feature, as it's nice to have commit id info in the image when someone runs the build.sh instead of Dockerfile directly.

CyberDem0n: `registry.opensource.zalan.do/acid/spilo-*` will have scm-source.json anyway, it is injected by CI/CD during the docker push

ReSearchITEng: done.

FxKu:
:+1:

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you @ReSearchITEng

CyberDem0n:
```
registry.opensource.zalan.do/acid/spilo-cdp-12:1.6-p114
```

PR #435, opened 2020-05-14
Added missing reference when to basebackup_fast_xlog replica method.

ermajn:
When not using WAL_E there was missing reference to basebackup replica method.
This is clear copy of #404


FxKu:
:+1:

CyberDem0n:
:+1:

Issue #413, opened 2020-02-18
Unable to build without Postgres 11 in PGOLDVERSIONS


gertvdijk:
Building master @ 84668553b9cda6ba9da199ff3b3fe9fd89cf5aec with

    $ ./build.sh [...] --build-arg PGOLDVERSIONS="" [...]

(Or anything that has no 11 in there.) Fails with:

```
+ PGVERSION=11
+ cd /usr/share/postgresql/11/contrib/postgis-3.0
/bin/sh: 1: cd: can't cd to /usr/share/postgresql/11/contrib/postgis-3.0
```

I believe that

https://github.com/zalando/spilo/blob/84668553b9cda6ba9da199ff3b3fe9fd89cf5aec/postgres-appliance/Dockerfile#L282

<strike>simply assumes that 11 is always included</strike> overrides the build-arg.

The context is that I'm trying to build an image that does not include all ancient versions of Postgres in an attempt to slim it down in size. I just want and need version 12.

I believe there's a bigger issue in those lines, because from that line on it just overrides the Docker build argument PGVERSION to 11, so the default of 12 is no longer 12 from that point. It's not obvious for me what is the solution, so I didn't create a PR (yet).

CyberDem0n:
The issue is that timescaledb is still not available for PG12, but we want to symlink all sql files it installs on different versions in order to make the image size a bit smaller.

DaniDD:
If I set PGVERSION to 10 and PGOLDVERSIONS="" then I become an error.
I think it need a check bevor you set the PGVERSION  to 11

aditya1808:
@gertvdijk does it work if we give PGVERSION and PGOLDVERSIONS  both as 11 ?? 
could you confirm this plz

gertvdijk:
> @gertvdijk does it work if we give PGVERSION and PGOLDVERSIONS both as 11 ??
> could you confirm this plz

Sorry, just noticing this now. Anyway, a PR with a fix is merged in the meantime. Should work. :)

Issue #433, opened 2020-05-06
issue when cloning https://github.com/CyberDem0n/pam-oauth2.git 


aditya1808:
Hi, im trying to build latest version 1.6-p3 in my company network using proxy.
ENV http_proxy=""
ENV https_proxy=""

I have provided proxy like mentioned above, in the Docker file.
But when building its failing when its trying to pull pam-oauth2 ,one of its submodules actually.

![image](https://user-images.githubusercontent.com/6763803/81187179-52e68f80-8fd1-11ea-8265-95cf9c450ef0.png)

i believe this submodule is using git:// instead of https://

Could you let me know how to fix this.

CyberDem0n:
If you change https://github.com/zalando/spilo/blob/master/postgres-appliance/Dockerfile#L77 to PAM_OAUTH2=v1.0.1 it should help

aditya1808:
thanks, that helped. I will close the ticket but i just want to know, is there any reason why its not implemented in spilo? ( **to use v1.0.1**)??

PR #438, opened 2020-05-19
Bump timescaledb version to 1.7.1

CyberDem0n:
In addition to that bump pam-oauth2 to v1.0.1, there is no real changes in the code, but it will solve https://github.com/zalando/spilo/issues/433

And last, fix build with `--build-arg PGOLDVERSIONS=""`

Close https://github.com/zalando/spilo/issues/413


avaczi:
:+1:

CyberDem0n:
:+1:

PR #454, opened 2020-06-26
Helper script for pgq

CyberDem0n:
We want to make it possible to grant pgq_* roles by mebers of admin role


erthalion:
:+1:

CyberDem0n:
:+1:

PR #475, opened 2020-08-05
Enable coredumps

erthalion:
To facilitate troubleshooting, enable coredumps. Unfortunately by
default a coredump will take snapshot of the whole shared memory of a
process, which in case of PostgreSQL could mean gigabyres of shared
buffers. To prevent that, also configure coredump_filter to exclude
shared memory from a dump. See [1] for more details.

[1]: https://www.kernel.org/doc/html/latest/filesystems/proc.html#proc-pid-coredump-filter-core-dump-filtering-settings

### DIFF
--- a/postgres-appliance/runit/patroni/run
+++ b/postgres-appliance/runit/patroni/run
@@ -11,6 +11,15 @@ if ! $CHPST true 2> /dev/null; then
     CHPST=""
 fi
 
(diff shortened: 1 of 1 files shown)


CyberDem0n:
:+1:

erthalion:
:+1:

PR #478, opened 2020-08-07
use AWS_REGION when provided

### LINKED ISSUE #477: AWS_REGION is not set when WAL_S3_BUCKET is not in the expected format.
When the value of WAL_S3_BUCKET is set to a string that does not conform to the pattern ${something}-${aws_region}-$-${something_else}, AWS_REGION fails to be parsed out https://github.com/zalando/spilo/blob/2113b5b7abda33c9f1a13a30d11ef106c67910d2/postgres-appliance/scripts/configure_spilo.py#L714 and is not set despite being originally provided in the environment variable AWS_REGION.

Under these circumstances would adding an `else` clause at line 720 to catch this case and add AWS_REGION back into the dictionary be appropriate?

```python
        # for S3-compatible storage we want to specify WALE_S3_ENDPOINT and AWS_ENDPOINT, but not AWS_REGION
        if aws_endpoint or wale_endpoint:
            if not aws_endpoint:
                aws_endpoint = wale_endpoint.replace('+path://', '://')
            elif not wale_endpoint:
                wale_endpoint = aws_endpoint.replace('://', '+path://')
            wale.update(WALE_S3_ENDPOINT=wale_endpoint, AWS_ENDPOINT=aws_endpoint, WALG_DISABLE_S3_SSE='true')
        elif not aws_region:
            # try to determine region from the endpoint or bucket name
            name = wale.get('WAL_S3_BUCKET') or wale.get('WALE_S3_PREFIX')
            match = re.search(r'.*(\w{2}-\w+-\d)-.*', name)
            if match:
                aws_region = match.group(1)
            else:
                aws_region = placeholders['instance_data']['zone'][:-1]
            wale['AWS_REGION'] = aws_region
        else:
            wale['AWS_REGION'] = aws_region
```

sweetbadger:
When AWS_REGION is supplied it is ignored when using AWS S3 storage
resolves #477

### DIFF
--- a/postgres-appliance/scripts/configure_spilo.py
+++ b/postgres-appliance/scripts/configure_spilo.py
@@ -717,6 +717,8 @@ def write_wale_environment(placeholders, prefix, overwrite):
             else:
                 aws_region = placeholders['instance_data']['zone'][:-1]
             wale['AWS_REGION'] = aws_region
+        else:
+            wale['AWS_REGION'] = aws_region
 
         if not (wale.get('AWS_SECRET_ACCESS_KEY') and wale.get('AWS_ACCESS_KEY_ID')):
             wale['AWS_INSTANCE_PROFILE'] = 'true'


Jan-M:
:+1:

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you @sweetbadger

PR #480, opened 2020-08-13
bugfix/issue479  Make sure created certificate are owned by postgres user

usernkey:
The owner of the certificates doesn't need to be customized  as there is only one possible option postgres.
This reduce that possibility that an user by mistake set a different user which cause the failure of the  bootstrap.


CyberDem0n:
:+1:

FxKu:
👍

PR #484, opened 2020-08-19
do not sum lists

yanchenko-igor:
We want to be able to overwrite default configuration.

### DIFF
--- a/postgres-appliance/scripts/configure_spilo.py
+++ b/postgres-appliance/scripts/configure_spilo.py
@@ -116,7 +116,7 @@ def deep_update(a, b):
     """Updates data structures
 
     Dicts are merged, recursively
-    List b is appended to a (except duplicates)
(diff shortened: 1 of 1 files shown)


postgres-appliance/scripts/configure_spilo.py
```diff
@@ -127,7 +127,7 @@ def deep_update(a, b):
                 a[key] = b[key]
         return a
     if type(a) is list and type(b) is list:
-        return a + [i for i in b if i not in a]
+        return a
```
avaczi, line 130: The function description needs to be adjusted accordignly

Jan-M: I agree

avaczi:
:+1:

CyberDem0n:
:+1:

PR #486, opened 2020-08-31
Changes configure script to use OpenStack metadata directly.

baylisscg:
Splits AWS and OpenStack metadata fetching.
Adds OpenStack specific fetching from http://169.254.169.254/openstack/latest/meta_data.json
Like GCP OpenStack's metadata does not expose IPv4 addresses so the auto-discovered one used.


Jan-M:
:+1:

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you @baylisscg

PR #487, opened 2020-09-03
Patroni 2.0

### LINKED ISSUE #468: Make etcd optional
Current version of Docker image is starting etcd regardless of if it's used or not. When using other dcs, etcd should not be started. @CyberDem0n Would you prefer a new env var to toggle this, or should it be derived by checking for presence of `ETCD_HOST`/`ETCD_HOSTS`/`ETCD_URL`?

CyberDem0n:
* wal-e 1.1.1
* wal-g 0.2.17
* timescaledb 1.7.3
* refactor DCS configuration (close #468)

### DIFF
--- a/postgres-appliance/Dockerfile
+++ b/postgres-appliance/Dockerfile
@@ -1,6 +1,6 @@
 ARG PGVERSION=12
-ARG TIMESCALEDB=1.7.2
-ARG TIMESCALEDB_LEGACY=1.7.2
+ARG TIMESCALEDB=1.7.3
+ARG TIMESCALEDB_LEGACY=1.7.3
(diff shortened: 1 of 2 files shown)


Jan-M:
:+1:

CyberDem0n:
:+1:

Legogris:
Am I correct in understanding that keeping `KUBERNETES_BYPASS_API_SERVICE` should change nothing and unless kubernetes is set as provider, no kubernetes/etcd-specific functionality is being expected now?

CyberDem0n:
It works only for cases when Patroni is using Kubernetes API for leader elections.
Also this parameters it is disabled by default.

PR #495, opened 2020-09-23
Disable prefetch when restore_command is called for pg_rewind

CyberDem0n:
We use the fact that `%p` parameter contains either `RECOVERYHISTORY` or `RECOVERYXLOG`, but in case of pg_rewind the file is being restored directly to `pg_wal/%f`
If it is detected, the following is done:
* export WALG_DOWNLOAD_CONCURRENCY=1 environment for wal-g
* Use -p 1 option for wal-e

In addition to that update README.rst and bump some dependencies.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -21,13 +21,14 @@ How to Build This Docker Image
 
     $ cd postgres-appliance
 
-    $ ./build.sh --build-arg COMPRESS=true --tag $YOUR_TAG .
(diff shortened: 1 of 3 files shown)


avaczi:
:+1:

sdudoladov:
:+1:

CyberDem0n:
:+1:

PR #510, opened 2020-10-20
'Initialization scripts' like in https://hub.docker.com/_/postgres implemented

yanchenko-igor:
This PR allows to add additional script after database initialization.

### DIFF
--- a/postgres-appliance/Dockerfile
+++ b/postgres-appliance/Dockerfile
@@ -1,5 +1,5 @@
 ARG PGVERSION=13
-ARG TIMESCALEDB=1.7.4
+ARG TIMESCALEDB=2.0.0
 ARG TIMESCALEDB_LEGACY=1.7.4
 ARG DEMO=false
 ARG COMPRESS=false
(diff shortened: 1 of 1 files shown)


docs/configuration.md
```diff
@@ -82,3 +82,26 @@ using KMS.
...
 in plain text anywhere. They will only be available to the EC2 instances running Spilo.
+
+### Initialization scripts
+
+If you would like to do additional initialization in an image derived from this one, add one or more `*.sql`, `*.sql.gz`, or `*.sh` scripts under `/docker-entrypoint-initdb.d` (creating the directory if necessary). After the entrypoint calls `initdb` to create the default `postgres` user and database, it will run any `*.sql` files, run any executable `*.sh` scripts, and source any non-executable `*.sh` scripts found in that directory to do further initialization before starting the service.
+
+**Warning**: scripts in `/docker-entrypoint-initdb.d` are only run if you start the container with a data directory that is empty; any pre-existing database will be left untouched on container startup. One common problem is that if one of your `/docker-entrypoint-initdb.d` scripts fails (which will cause the entrypoint script to exit) and your orchestrator restarts the container with the already initialized data directory, it will not continue on with your scripts.
```
CyberDem0n, line 90: That's not entirely correct.
Patroni executes post_init.sh when:
1. initializing the new cluster (initdb or custom bootstrap)
2. On promote

docs/configuration.md
```diff
@@ -82,3 +82,26 @@ using KMS.
...
+
+Additionally, as of [docker-library/postgres#253](https://github.com/docker-library/postgres/pull/253), these initialization scripts are run as the `postgres` user (or as the "semi-arbitrary user" specified with the `--user` flag to `docker run`; see the section titled "Arbitrary `--user` Notes" for more details). Also, as of [docker-library/postgres#440](https://github.com/docker-library/postgres/pull/440), the temporary daemon started for these initialization scripts listens only on the Unix socket, so any `psql` usage should drop the hostname portion (see [docker-library/postgres#474 (comment)](https://github.com/docker-library/postgres/issues/474#issuecomment-416914741) for example).
```
CyberDem0n, line 107: Neither `POSTGRES_USER` nor `POSTGRES_DB` are set by default.

The post_init.sh is executed with postgres user. Also post_init.sh receives two arguments, the first one is $HUMAN_ROLE and the second - connection string to postgres database.

postgres-appliance/Dockerfile
```diff
@@ -482,6 +482,7 @@ ENV WALE_ENV_DIR=$RW_DIR/etc/wal-e.d/env \
 ENV PGDATA=$PGROOT/data \
     PGLOG=$PGROOT/pg_log
 
+WORKDIR /docker-entrypoint-initdb.d
```
CyberDem0n, line 485: I don't think that we need this one.

postgres-appliance/scripts/post_init.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -Eeo pipefail
```
CyberDem0n, line 2: Please remove it.
There are places where commands could exit with non-zero exit code (it is expected). We don't want the script to break because of that.

CyberDem0n (changes requested):

Issue #346, opened 2019-07-05
Build-arg should support Locale


schmitch:
at the moment the image can only run as en_us, however it would be great if it would be possible to also specify the local as a build_arg

schmitch:
closing, as @fraanek made a really really good MR for that! thanks!

PR #513, opened 2020-11-13
Support for additional locales

fraanek:
Support for additional UTF-8 locales build into image by specify build-arg - #346

You can specify **ADDITIONAL_LOCALES** build-arg as comma separated list of locales to build into image (example: _de_DE,pl_PL_). You can also specify environment variable **INITDB_LOCALE** when running image as container to change database cluster's default locale.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -31,6 +31,7 @@ There are a few build arguments defined in the Dockerfile and it is possible to 
 - PGOLDVERSIONS="9.5 9.6 10 11"
 - DEMO=false # set to true to build the smallest possible image which will work only on Kubernetes
 - TIMESCALEDB_APACHE_ONLY=true # set to false to build timescaledb community version (Timescale License)
+- ADDITIONAL_LOCALES= # comma separated locales to build into image (example: de_DE,pl_PL)
 
 Run the image locally after build:
 
--- a/README.rst
+++ b/README.rst
@@ -31,7 +31,7 @@ There are a few build arguments defined in the Dockerfile and it is possible to 
 - PGOLDVERSIONS="9.5 9.6 10 11"
 - DEMO=false # set to true to build the smallest possible image which will work only on Kubernetes
 - TIMESCALEDB_APACHE_ONLY=true # set to false to build timescaledb community version (Timescale License)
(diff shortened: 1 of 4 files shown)


README.rst
```diff
@@ -31,6 +31,7 @@ There are a few build arguments defined in the Dockerfile and it is possible to
 - PGOLDVERSIONS="9.5 9.6 10 11"
 - DEMO=false # set to true to build the smallest possible image which will work only on Kubernetes
 - TIMESCALEDB_APACHE_ONLY=true # set to false to build timescaledb community version (Timescale License)
+- ADDITIONAL_LOCALES= # comma separated locales to build into image (example: de_DE,pl_PL)
```
CyberDem0n, line 34: I wonder if we should be using a comma as a separator, just space character (like it is used in the `PGOLDVERSIONS`) would work nice and will make `echo | sed` call unnecessary:
https://github.com/zalando/spilo/pull/513/files#diff-8918d403d0c48088cebb6ed78df4fecd856e211f7522e06ee423383b17bd8453R32
https://github.com/zalando/spilo/pull/513/files#diff-8918d403d0c48088cebb6ed78df4fecd856e211f7522e06ee423383b17bd8453R44

postgres-appliance/Dockerfile
```diff
@@ -3,9 +3,12 @@ ARG TIMESCALEDB=1.7.4
 ARG TIMESCALEDB_LEGACY=1.7.4
 ARG DEMO=false
 ARG COMPRESS=false
+ARG ADDITIONAL_LOCALES=
```
CyberDem0n, line 6:
```suggestion
```
Global ARG is only required when we want to use it in multiple stages.

CyberDem0n (changes requested):

README.rst
fraanek: Changed comma to space character

postgres-appliance/Dockerfile
fraanek: Removed global ARG

CyberDem0n (approved):

Jan-M:
:+1:

Jan-M (approved):

CyberDem0n:
👍

PR #518, opened 2020-12-04
Fix clone with upgrade

CyberDem0n:
Due to the rename of wal_keep_segments to wal_keep_size in pg13 clone with upgrade stopped working because the maybe_upgrade scripts tries to start postgres (using patroni libraries) while config is not adapted to the new version.
The solution is calling configure_server_parameters() after performing a switch of PGDATA.

To avoid future similar issues added the test case of upgrade 9.5->13

Last, but not least, finally fixed behavior of `bin_dir`.
The following order of precedence is used:
* $PGDATA/PG_VERSION
* $SPILO_CONFIGURATION
* $PGVERSION


FxKu:
:+1:

CyberDem0n:
:+1:

Issue #511, opened 2020-10-29
Replicas fail regularly when /etc/service dissapears


jf647:
I'm running Spilo via postgres-operator.  I have a cluster of 3 pods on a GKE Kubernetes cluster.  Every 24-36 hours, one or both of the replicas fails.  If I tail the pod logs, I see this over and over again:

```
default/nadt-pgsql-2[postgres]: runsvdir /etc/service: warning: unable to stat /etc/service: file does not exist
```

And indeed, if I check the pod, /run/service (which /etc/service is linked to) is gone:

```
$ kubectl exec -i -t -n default nadt-pgsql-0 -- ls -la /etc/service
lrwxrwxrwx 1 root root 12 Apr 24  2020 /etc/service -> /run/service
$ kubectl exec -i -t -n default nadt-pgsql-0 -- ls -la /run/service
ls: cannot access '/run/service': No such file or directory
command terminated with exit code 2
```

Immediately preceeding the runsvdir errors in the log, I get a series of uncaught exceptions, which I've copied in this gist:

https://gist.github.com/jf647/b5404f6102b0917f1db43f3e76afce3e

My pods are running the registry.opensource.zalan.do/acid/spilo-12:1.6-p3 image and my postres-operator is running registry.opensource.zalan.do/acid/postgres-operator:v1.5.0.

Deleting the affected pods and allowing them to bootstrap from the master seems to clear things up, but that's toil and I'm hoping to get to the root of how the containers get into this state.  I'm not sure what other data to collect, but as it happens so often I can grab anything needed the next time it happens.

jf647:
Should this have been opened in zalando/postgres-operator instead?  I'm having to restart containers every day (because the pod template doesn't allow a sidecar definition), and if it happens in the middle of the night (this is a personal DB providing no critical services, and thus doesn't page 24/7), all of the pods have usually failed and I have to restore from backup.

CyberDem0n:
You need to figure out what wipes out /run directory.
There's nothing in spilo what could do that.

jf647:
As the gist shows, this is an uncaught Python exception when configure_spilo.py is running.

It appears that the script fails to gracefully degrade if the DNS lookup for the metadata server fails:

https://github.com/zalando/spilo/blob/85da543f8a5eb64e26cb67e248e9dee22091ee09/postgres-appliance/scripts/configure_spilo.py#L313

```
Max retries exceeded with url: /computeMetadata/v1/instance/zone
```

```
[...]

PR #521, opened 2020-12-15
Use an IP address for Google instance metadata

alexeyklyukin:
Makes Spilo start even if external DNS queries from
within the container are not working.

Do not continue with the startup when /etc/service is missing.

Per https://github.com/zalando/spilo/issues/511

### DIFF
--- a/postgres-appliance/scripts/configure_spilo.py
+++ b/postgres-appliance/scripts/configure_spilo.py
@@ -376,7 +376,7 @@ def get_instance_metadata(provider):
     headers = {}
     if provider == PROVIDER_GOOGLE:
         headers['Metadata-Flavor'] = 'Google'
-        url = 'http://metadata.google.internal/computeMetadata/v1/instance'
+        url = 'http://169.254.169.254/computeMetadata/v1/instance' # metadata.google.internal
         mapping = {'zone': 'zone'}
         if not USE_KUBERNETES:
             mapping.update({'id': 'id'})
--- a/postgres-appliance/scripts/configure_spilo.py
+++ b/postgres-appliance/scripts/configure_spilo.py
@@ -376,7 +376,7 @@ def get_instance_metadata(provider):
     headers = {}
     if provider == PROVIDER_GOOGLE:
         headers['Metadata-Flavor'] = 'Google'
-        url = 'http://169.254.169.254/computeMetadata/v1/instance' # metadata.google.internal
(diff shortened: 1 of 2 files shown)


postgres-appliance/scripts/configure_spilo.py
```diff
@@ -376,7 +376,7 @@ def get_instance_metadata(provider):
     headers = {}
     if provider == PROVIDER_GOOGLE:
         headers['Metadata-Flavor'] = 'Google'
-        url = 'http://metadata.google.internal/computeMetadata/v1/instance'
+        url = 'http://169.254.169.254/computeMetadata/v1/instance' # metadata.google.internal
```
CyberDem0n, line 379: Flake8 complains: postgres-appliance/scripts/configure_spilo.py:379:67: E261 at least two spaces before inline comment
```suggestion
        url = 'http://169.254.169.254/computeMetadata/v1/instance'  # metadata.google.internal
```

CyberDem0n (changes requested):
Now we have some tests and basic linting of `py` and `sh` files. Please make it green.

CyberDem0n (approved):

CyberDem0n:
:+1:

postgres-appliance/scripts/configure_spilo.py
```diff
@@ -376,7 +376,7 @@ def get_instance_metadata(provider):
     headers = {}
     if provider == PROVIDER_GOOGLE:
         headers['Metadata-Flavor'] = 'Google'
-        url = 'http://metadata.google.internal/computeMetadata/v1/instance'
+        url = 'http://169.254.169.254/computeMetadata/v1/instance'  # metadata.google.internal
```
Jan-M, line 379: should this be env var?

Jan-M:
I am giving +1 cannot assess if it should be env var with default ip or dns name. Might never change though, but could come in handy.

Jan-M:
:+1:

postgres-appliance/scripts/configure_spilo.py
alexeyklyukin: I thought about it, but decided against: to me it seems readability, being able to see the whole url at once, trumps the benefits of defining the base url in a one place only in this case, especially given that the address is very unlikely to change.

alexeyklyukin:
@CyberDem0n @Jan-M thank you for reviewing and merging it!

PR #532, opened 2021-01-03
Dockerfile: enable full Timescale features

haf:
This changes the license that the TimescaleDB (that has to explicitly be enabled by the user) ships with to the [Timescale license](https://www.timescale.com/legal/licenses).

See https://github.com/zalando/spilo/pull/419#issuecomment-705646944 for reasoning.

Alternatively, would you consider supporting a build on your infrastructure that toggles this flag off, as a tag of the Spilo images that you make available?


postgres-appliance/Dockerfile
```diff
@@ -70,7 +70,7 @@ COPY dependencies/debs dependencies/src /builddeps/
 ARG PGVERSION
 ARG TIMESCALEDB
 ARG TIMESCALEDB_LEGACY
-ARG TIMESCALEDB_APACHE_ONLY=true
+ARG TIMESCALEDB_APACHE_ONLY=false
```
CyberDem0n, line 73: I am fine with the change of the default build argument in the Dockerfile, so everyone who builds the image on their own will automatically get Timescale with TSL.

But, we don't want to take responsibility for building and making publicly available the image with TSL license. Therefore please also update the [delivery.yaml](https://github.com/zalando/spilo/blob/master/delivery.yaml#L27) to
```
docker build --build-arg PGVERSION=$PGVERSION --build-arg TIMESCALEDB_APACHE_ONLY=true -t $IMAGE .
```

Also, it would be nice to reflect the change of the TIMESCALEDB_APACHE_ONLY default value in the [README](https://github.com/zalando/spilo/blob/master/README.rst#how-to-build-this-docker-image) file.

haf: Thank you for your reply and considering changing it.

Would you be open to two build tags, one with TSL and one without?

CyberDem0n: I afraid we will not take a responsibility on distributing the software with non open-source (TSL) license.

haf: Ok. Thanks for meeting me half way, even so.

haf:
I've done the updates you requested.

Jan-M:
I do not see value in toggling our default? We allow already to overwrite if you want to build with a different config?

patrykwojtynski:
Hey, is there any reason to not have it toggled by default? Using it in combination with zalando/postgres-operator would be easier, especially in terms of upgrading between this operator releases.

kuzaxak:
> Hey, is there any reason to not have it toggled by default? Using it in combination with zalando/postgres-operator would be easier, especially in terms of upgrading between this operator releases.

I think the main reason was described [here][1]:

> I afraid we will not take a responsibility on distributing the software with non open-source (TSL) license.

[1]: https://github.com/zalando/spilo/pull/532#discussion_r552420704

Jan-M:
Closing, as per earlier comment which describes our policy to not enable those feature in the Spilo image.

Issue #528, opened 2020-12-20
2.0-p2 can not build image


necipakca:
I can not build 2.0-p2 release with additional language.

Build process exit with code 2.

+ cd /usr/share/postgresql/12/contrib/postgis-3.0
/bin/sh: 1: cd: can't cd to /usr/share/postgresql/12/contrib/postgis-3.0

I added all build process log to attachment. 

[spilo_build.log](https://github.com/zalando/spilo/files/5721262/spilo_build.log)

CyberDem0n:
This is due to the postgis 3.1 release.

aditya1808:
How to resolve this now?

mindw00rk:
Dockerfile:
ENV POSTGIS_VERSION=3.0 -> ENV POSTGIS_VERSION=3.1
${POSTGIS_VERSION%.0} -> ${POSTGIS_VERSION%.1}

PR #534, opened 2021-01-03
Update Dockerfile

haf:
Support PostGIS 3.1

Fixes #529

Issue #533, opened 2021-01-03
Upgrade to Timescale 2.0


haf:
Timescale 2.0 was released 13 days ago https://github.com/timescale/timescaledb/releases/tag/2.0.0 — it works well on top of ordinary pgsql from my testing, but has some differences in how it manages policies and compression. I think this version should be the default in the spilo image.

- [ ] Upgrade notes https://docs.timescale.com/latest/update-timescaledb/update-tsdb-2
- [ ] You'll need cmake 3.11 or higher. Current release is 3.19 https://graspingtech.com/upgrade-cmake/
   - [ ] Alternatively use [Ubuntu 20.04 LTS]( https://wiki.ubuntu.com/Releases) that ships `cmake version 3.16.3`
   - [ ] Release notes https://docs.timescale.com/latest/release-notes
- [ ] Consider upgrading PostGIS at the same time https://github.com/zalando/spilo/issues/529
- [ ] You need to remain on pg 12 until https://github.com/timescale/timescaledb/pull/2498 is merged

CMake quickfix:

```Dockerfile
 # install modern cmake
RUN curl -o cmake.tar.gz -LO https://github.com/Kitware/CMake/releases/download/v${CMAKE_VERSION}/cmake-${CMAKE_VERSION}-Linux-x86_64.tar.gz \
        && tar -zxf cmake.tar.gz \
        && mv cmake-${CMAKE_VERSION}-Linux-x86_64/bin/ /usr/local/bin
```

Works well because `/usr/local/bin` is normally empty and the binary is compiled for the "normal" x86_64 arch already.

haf:
```Dockerfile
ARG PGVERSION=12
ARG TIMESCALEDB=2.0.0
ARG TIMESCALEDB_LEGACY=2.0.0
ARG DEMO=false
ARG COMPRESS=false

FROM ubuntu:18.04 as builder-false

ARG DEMO
ARG ADDITIONAL_LOCALES=

RUN export DEBIAN_FRONTEND=noninteractive \
    && echo 'APT::Install-Recommends "0";\nAPT::Install-Suggests "0";' > /etc/apt/apt.conf.d/01norecommend \
    && apt-get update \
    && apt-get install -y curl ca-certificates less locales jq vim-tiny gnupg1 cron runit dumb-init libcap2-bin rsync \
    && ln -s chpst /usr/bin/envdir \
    # Make it possible to use the following utilities without root
    && setcap 'cap_sys_nice+ep' /usr/bin/chrt \
[...]

Issue #529, opened 2020-12-21
Upgrade to PostGIS 3.1.0


boonen:
On 18 December 2020 PostGIS 3.1.0 was released. This version has several important improvements with regard to performance, so I would like to see it included in the next Spilo release. For details see this blog post: <https://postgis.net/2020/12/18/postgis-3.1.0/>.

If you like I am willing to provide a pull request.

CyberDem0n:
Closed in https://github.com/zalando/spilo/pull/535

boonen:
@CyberDem0n Thanks for solving this issue so quickly!

PR #535, opened 2021-01-04
Postgis 3.1, timescale 2.0 and a few optimizations in relinking of extensions/contribs

CyberDem0n:
1. Fix build broken by postgis 3.1 release
   - use postgis 3.1 as main version
   - use postgis 3.0 as legacy version (postgres 9.5)
   - fix relinking code to support legacy postgis and optimize it to support relinking between all major versions, not only the current one.
2.  Bump version of timescaledb to 2.0
    - change cmake requirement from 3.11 to 3.11 in CMakeLists.txt. The old version available in ubuntu 18.04 perfectly works.
    - call bootstrap with -DWARNINGS_AS_ERRORS=OFF
    - 9.6 and 10 will continue using 1.7.4, therefore optimized relinking code in the commit updating postgis becomes very handy.
3. A bit unrelated, but since this PR already does a lot of housekeeping, optimize the code defining/cleaning locales in order to reduce the number of repetitions.

Close: #528, #529, #534, #533

### DIFF
--- a/postgres-appliance/Dockerfile
+++ b/postgres-appliance/Dockerfile
@@ -79,7 +79,8 @@ ARG WITH_PERL=false
 ARG DEB_PG_SUPPORTED_VERSIONS="$PGOLDVERSIONS $PGVERSION"
 
 # Install PostgreSQL, extensions and contribs
-ENV POSTGIS_VERSION=3.0 \
+ENV POSTGIS_VERSION=3.1 \
+    POSTGIS_LEGACY=3.0 \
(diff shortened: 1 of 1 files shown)


FxKu:
:+1:

CyberDem0n:
:+1:

haf:
How does this address the fact that TS 2 doesn't support PG 13?

CyberDem0n:
Pg 13 in spilo is just running without timescale.

PR #538, opened 2021-01-08
'Initialization scripts' like in https://hub.docker.com/_/postgres implemented

yanchenko-igor:


postgres-appliance/scripts/post_init.sh
```diff
@@ -158,5 +160,55 @@ GRANT EXECUTE ON FUNCTION public.set_user(text) TO admin;
...
+
+	"${query_runner[@]}" "$@"
+}
+
+
+
+# check dir permissions to reduce likelihood of half-initialized database
+ls /docker-entrypoint-initdb.d/ > /dev/null
+
+docker_process_init_files /docker-entrypoint-initdb.d/*
```
CyberDem0n, line 214: Since in the readme you [tell](https://github.com/zalando/spilo/pull/538/files#diff-17ed18489a956f326ec0fe4040850c5bc9261d4631fb42da4c52891d74a59180R90) that scripts will be executed only for the new cluster, it makes sense to execute this block only if we know that post_init is not executed from on_role_change:

```bash
if [ "$2" != "postgres" ]; then
    ...
fi
```

CyberDem0n (changes requested):

Issue #542, opened 2021-01-14
Issues with Azure Kubernetes Service with AAD Pod Identity


keyz182:
I'm using the `registry.opensource.zalan.do/acid/spilo-13:2.0-p2` image via the postgres operator. It appears that when using [AAD Pod Indentity](https://github.com/Azure/aad-pod-identity) on an Azure Kubernetes Service instance, some of the provider detection logic is failing. If I override `SPILO_PROVIDER` to `local` things work. 

```
2021-01-14T16:28:17.428933191Z 2021-01-14 16:28:17,428 - bootstrapping - INFO - Figuring out my environment (Google? AWS? Openstack? Local?)
2021-01-14T16:28:17.443798096Z Traceback (most recent call last):
2021-01-14T16:28:17.443810896Z   File "/scripts/configure_spilo.py", line 1012, in <module>
2021-01-14T16:28:17.443814096Z     main()
2021-01-14T16:28:17.443820197Z   File "/scripts/configure_spilo.py", line 890, in main
2021-01-14T16:28:17.443823897Z     placeholders = get_placeholders(provider)
2021-01-14T16:28:17.443827297Z   File "/scripts/configure_spilo.py", line 598, in get_placeholders
2021-01-14T16:28:17.443830697Z     placeholders['instance_data'] = get_instance_metadata(provider)
2021-01-14T16:28:17.443833797Z   File "/scripts/configure_spilo.py", line 391, in get_instance_metadata
2021-01-14T16:28:17.443837298Z     openstack_metadata = requests.get(url, timeout=5).json()
2021-01-14T16:28:17.443839498Z   File "/usr/lib/python3/dist-packages/requests/models.py", line 884, in json
2021-01-14T16:28:17.443841798Z     self.content.decode(encoding), **kwargs
2021-01-14T16:28:17.443843998Z   File "/usr/lib/python3.6/json/__init__.py", line 354, in loads
2021-01-14T16:28:17.443846198Z     return _default_decoder.decode(s)
2021-01-14T16:28:17.443848298Z   File "/usr/lib/python3.6/json/decoder.py", line 339, in decode
2021-01-14T16:28:17.443850498Z     obj, end = self.raw_decode(s, idx=_w(s, 0).end())
2021-01-14T16:28:17.443852598Z   File "/usr/lib/python3.6/json/decoder.py", line 357, in raw_decode
2021-01-14T16:28:17.443854798Z     raise JSONDecodeError("Expecting value", s, err.value) from None
[...]

PR #543, opened 2021-01-15
[issues/542] Try to parse response to ensure it's parsable. If not, assume local provider

keyz182:
For issue #542 

Calls `r.json()` to try to parse the response, and on `JSONDecodeError` returns `PROVIDER_LOCAL` instead of `PROVIDER_OPENSTACK`.

### DIFF
--- a/postgres-appliance/scripts/configure_spilo.py
+++ b/postgres-appliance/scripts/configure_spilo.py
@@ -6,6 +6,8 @@ import json
 import logging
 import re
 import os
+from json.decoder import JSONDecodeError
+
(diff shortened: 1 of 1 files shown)


keyz182:
The actual error I encountered in #542 was in the `get_instance_metadata` method, but placing this check in `get_provider` instead seemed more appropriate, as it can return the local provider immediately, and help prevent further errors from assuming it's openstack.

avaczi:
:+1:

CyberDem0n:
:+1:

PR #545, opened 2021-01-21
Run CHECKPOINT on replicas in parallel with pg_upgrade

CyberDem0n:
Major upgrade with replica rsync requires all replicas being up-to-date with the primary and shutdown cleanly. It is possible to make the shutdown process faster by executing the CHECKPOINT immediately before running `pg_ctl stop -m fast`.

Before this commit, the CHECKPOINT and shutdown on replicas were performed only after the pg_upgrade was successfully executed.

The time for CHECKPOINT on replicas and therefore time for the shutdown is quite measurable and comparable with the time which is required to execute the pg_upgrade.
By doing these processes (pg_upgrade and CHECKPOINT on replicase) in parallel we reduce the downtime that is required for pg_upgrade + rsync.

The old diagram:
```
                                +-> r1 CHECKPOINT +   +-> r1 rsync +
SHUTDOWN PRIMARY -> pg_upgrade -|-> r2 CHECKPOINT |->-|-> r2 rsync |->
                                +-> r3 CHECKPOINT +   +-> r3 rsync +
```

The new diagram:
```
                  +-> pg_upgrade    +
                  |-> r1 CHECKPOINT |   +-> r1 rsync +
SHUTDOWN PRIMARY -|-> r2 CHECKPOINT |->-|-> r2 rsync |->
                  +-> r3 CHECKPOINT +   +-> r3 rsync +

```

In addition to that fix corner-case with upgrading a single-node cluster.


sdudoladov:
:+1:

CyberDem0n:
:+1:

PR #546, opened 2021-01-22
Set AWS_REGION whenever USE_WALG_BACKUP is set to true

vfrans:
Since wal-g supports having both AWS_ENDPOINT (or WALE_S3_ENDPOINT) and AWS_REGION and that we cannot determine region from endpoint in the event of an onpremise deployment, this PR allows setting AWS_REGION even if AWS_ENDPOINT (or WALE_S3_ENDPOINT) is specified.

This way using an onpremise deploy with a custom region (http://someminioinstance:9000/ and my-place, ie) or using an endpoint that does not match the regex engine (https://somebucket.s3.nl-ams.scw.cloud, ie)

### DIFF
--- a/postgres-appliance/scripts/configure_spilo.py
+++ b/postgres-appliance/scripts/configure_spilo.py
@@ -719,7 +719,7 @@ def write_wale_environment(placeholders, prefix, overwrite):
             elif not wale_endpoint:
                 wale_endpoint = aws_endpoint.replace('://', '+path://')
             wale.update(WALE_S3_ENDPOINT=wale_endpoint, AWS_ENDPOINT=aws_endpoint, WALG_DISABLE_S3_SSE='true')
-            if wale.get('USE_WALG_BACKUP') and wale.get('USE_WALG_BACKUP') == 'true':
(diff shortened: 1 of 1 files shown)


postgres-appliance/scripts/configure_spilo.py
```diff
@@ -719,6 +719,8 @@ def write_wale_environment(placeholders, prefix, overwrite):
             elif not wale_endpoint:
                 wale_endpoint = aws_endpoint.replace('://', '+path://')
             wale.update(WALE_S3_ENDPOINT=wale_endpoint, AWS_ENDPOINT=aws_endpoint, WALG_DISABLE_S3_SSE='true')
+            if wale.get('USE_WALG_BACKUP') and wale.get('USE_WALG_BACKUP') == 'true':
```
CyberDem0n, line 722: We want to be sure that aws_region is actually defined:
```suggestion
            if aws_region and wale.get('USE_WALG_BACKUP') == 'true':
```

sdudoladov:
:+1:

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you @vfrans!

Issue #548, opened 2021-01-29
Support for Wal-G over SSH transport


Spindel:
Since pre-release v0.2.17, and now in stable release v0.2.19,  Wal-G has support for using ssh as the transport as well as S3 and others.

It would be neat to be able to use this for the use-cases where it's a bit annoying  to set up an S3 compatible storage layer.

PR #549, opened 2021-01-29
Bump the Wal-G version to 0.2.19 and add WALG  SSH support

Spindel:
Bump the Wal-G version to 0.2.19 and add variables for the WALG_SSH_PREFIX for backups

Not adding SSH_PASSWORD as password authentication is generally not recommended.

The SSH transport functions using sftp://, meaning that the server side could
be set up in a completely restricted manner in sshd_config:

    Match User spilo:
        ChrootDirectory /srv/backup/spilo
        ForceCommand internal-sftp
        AuthenticationMethods publickey
        PermitTunnel no
        AllowAgentForwarding no
        AllowTcpForwarding no
        X11Forwarding no

With the matching setup in the spilo container could be:

SSH_PRIVATE_KEY_PATH=/etc/patroni/ssh_key
SSH_USER=patroni
WALG_SSH_PREFIX=sftp://backup.example.com/spilo

Fixes: zalando/spilo#548

### DIFF
--- a/postgres-appliance/scripts/configure_spilo.py
+++ b/postgres-appliance/scripts/configure_spilo.py
@@ -753,7 +753,7 @@ def write_wale_environment(placeholders, prefix, overwrite):
     elif wale.get("WALG_AZ_PREFIX"):
         write_envdir_names = azure_names + walg_names
     elif wale.get("WALG_SSH_PREFIX"):
-        wale_envdir_names = ssh_names + walg_names
+        write_envdir_names = ssh_names + walg_names
     else:
         return
 
--- a/postgres-appliance/scripts/configure_spilo.py
+++ b/postgres-appliance/scripts/configure_spilo.py
@@ -34,7 +34,7 @@ USE_KUBERNETES = os.environ.get('KUBERNETES_SERVICE_HOST') is not None
 KUBERNETES_DEFAULT_LABELS = '{"application": "spilo"}'
 MEMORY_LIMIT_IN_BYTES_PATH = '/sys/fs/cgroup/memory/memory.limit_in_bytes'
 PATRONI_DCS = ('zookeeper', 'exhibitor', 'consul', 'etcd3', 'etcd')
-AUTO_ENABLE_WALG_RESTORE = ('WAL_S3_BUCKET', 'WALE_S3_PREFIX', 'WALG_S3_PREFIX', 'WALG_AZ_PREFIX')
+AUTO_ENABLE_WALG_RESTORE = ('WAL_S3_BUCKET', 'WALE_S3_PREFIX', 'WALG_S3_PREFIX', 'WALG_AZ_PREFIX', 'WALG_SSH_PREFIX')
 WALG_SSH_NAMES = ['WALG_SSH_PREFIX', 'SSH_PRIVATE_KEY_PATH', 'SSH_USERNAME']
 
 
@@ -576,7 +576,7 @@ def get_placeholders(provider):
 
     placeholders['USE_WALE'] = any(placeholders.get(n) for n in AUTO_ENABLE_WALG_RESTORE +
                                    ('WAL_SWIFT_BUCKET', 'WALE_SWIFT_PREFIX', 'WAL_GCS_BUCKET',
-                                    'WAL_GS_BUCKET', 'WALE_GS_PREFIX', 'WALG_GS_PREFIX', 'WALG_SSH_PREFIX'))
+                                    'WAL_GS_BUCKET', 'WALE_GS_PREFIX', 'WALG_GS_PREFIX'))
 
     if placeholders.get('WALG_BACKUP_FROM_REPLICA'):
         placeholders['WALG_BACKUP_FROM_REPLICA'] = str(placeholders['WALG_BACKUP_FROM_REPLICA']).lower()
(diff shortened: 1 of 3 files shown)


postgres-appliance/scripts/configure_spilo.py
```diff
@@ -569,11 +570,13 @@ def get_placeholders(provider):
...
     if placeholders.get('WALG_AZ_PREFIX'):
         placeholders.setdefault('USE_WALG_BACKUP', 'true')
+    if all(placeholders.get(n) for n in WALG_SSH_NAMES):
+        placeholders.setdefault('USE_WALG_BACKUP', 'true')
     set_walg_placeholders(placeholders)
 
     placeholders['USE_WALE'] = any(placeholders.get(n) for n in AUTO_ENABLE_WALG_RESTORE +
                                    ('WAL_SWIFT_BUCKET', 'WALE_SWIFT_PREFIX', 'WAL_GCS_BUCKET',
-                                    'WAL_GS_BUCKET', 'WALE_GS_PREFIX', 'WALG_GS_PREFIX'))
+                                    'WAL_GS_BUCKET', 'WALE_GS_PREFIX', 'WALG_GS_PREFIX', 'WALG_SSH_PREFIX'))
```
CyberDem0n, line 579: Please add `WALG_SSH_PREFIX` to the [AUTO_ENABLE_WALG_RESTORE](https://github.com/zalando/spilo/pull/549/files#diff-e4dcf964214d58d733ae9874d4a2ab4b8e9eb99213307e1666af5dad21a326bbR37).

And here it could be removed.

CyberDem0n (changes requested):
Hi @Spindel,

thank you for the PR.
There is a small change required to make it better.
Have you actually already tried to use the image with this change?

Also, regarding the new version of wal-g. Not so long time ago we had some issues with 0.2.17 working with AWS S3 (but no problems with minio). Therefore we had to roll it back to 0.2.15. I really hope 0.2.19 will work fine.

postgres-appliance/scripts/configure_spilo.py
Spindel: As I haven't tested it with the AUTO features at all yet, I left it unset. I honestly find spilo_configure to be quite obtuse and difficult to parse.

As I don't use S3 I cannot speak for if it works there or not with this change.  Tests of backing up have been done, but only on small data sets as I'm currently evaluating spilo for our environment, and backing up over SSH is one of our production requirements, so I needed to get this fixed and working first.

I'll follow up with changes as requested once I've tested them, probably later Tuseday.

CyberDem0n: > As I haven't tested it with the AUTO features at all yet, I left it unset. I honestly find spilo_configure to be quite obtuse and difficult to parse.

Sorry about that. It was "growing" for 6 years with mostly ad-hoc changes. That pretty much explains the current state.

> As I don't use S3 I cannot speak for if it works there or not with this change.

That's fine. We have all the tests. First of all intergartion tests with docker-compose: https://github.com/zalando/spilo/tree/master/postgres-appliance/tests, which are execute for every commit, but Minio is used. Also we have end-to-end tests on a real K8s cluster running on AWS. The experience was that end-to-end tests were failing due to the wal-g 0.2.17  not able to communicate with S3, while 0.2.15 was working perfectly fine. If this behavior retain I afraid we can't use the latest version :(

> Tests of backing up have been done

It also would be nice to do a restore, otherwise we are dealing with Schrodinger backup. Actualy this is one of the point of having automated tests where we check different scanarios including backup is created and it is possible to restore from it.

> I'll follow up with changes as requested once I've tested them, probably later Tuseday.

Cool, thank you!

Spindel: Followup:

Testing in progress with the changes requested. I got side-tracked by a build failure in CI when trying to make the container build for us, turns out it was an unrelated capability difference. 

Running a build set and tests. Previous commit at least showed up as working for me.

Unrelated question, but the sysstat package is being installed in the spilo images, and automatically triggered via cron when enabling the backup feature. Is that really intended, or some side-effect of the debian packaging?

Spindel: And now I can confirm that with the requested change ( 9c5c8e1cecda671eab3d78a70869768677111a8b )  it works and can automatically restore  with wal-g as well.

CyberDem0n (approved):

sdudoladov:
:+1:

sdudoladov (approved):

CyberDem0n:
:+1:

postgres-appliance/scripts/configure_spilo.py
CyberDem0n: > Unrelated question, but the sysstat package is being installed in the spilo images, and automatically triggered via cron when enabling the backup feature. Is that really intended, or some side-effect of the debian packaging?

The `sysstat` is installed only just for convenience, for example one could run `iostat -mx 1` from the container.

Backups are triggered from the cron, yes, but it is not related to the sysstat.

Spindel: Ah, right. Mind if I file a change to turn sysstat on/off as a build argument?

CyberDem0n: We only install a package, but don't start the daemon.
Package itself is quite small, less than 1MB. On top of 720MB this is peanuts.

Spindel:
Yea, it wasn't about size, but rather about the change in pid count and
disk usage inside the container that triggered  something here when the
sysstat started to run in cron just because I enabled backups. Not a huge
problem, but it was an unknown side-effect of enabling backups.

Issue #551, opened 2021-02-05
seteuid: Operation not permitted Error in PG cluster


neelasha-09:
Hi,

We see below error while running postgres-operator-cluster  on OpenShift environment.

Image: `spilo-13:2.0-p3`

```shell
[crc@eaas-crc postgres-operator]$ oc logs -f postgres-service-01-0 -c postgres
2021-02-05 11:23:06,997 - bootstrapping - INFO - Figuring out my environment (Google? AWS? Openstack? Local?)
2021-02-05 11:23:08,089 - bootstrapping - INFO - Could not connect to 169.254.169.254, assuming local Docker setup
2021-02-05 11:23:08,091 - bootstrapping - INFO - No meta-data available for this provider
2021-02-05 11:23:08,091 - bootstrapping - INFO - Looks like your running local
2021-02-05 11:23:08,168 - bootstrapping - INFO - Configuring crontab
2021-02-05 11:23:08,169 - bootstrapping - INFO - Skipping creation of renice cron job due to lack of SYS_NICE capability
2021-02-05 11:23:08,186 - bootstrapping - INFO - Configuring pgqd
2021-02-05 11:23:08,187 - bootstrapping - INFO - Configuring log
2021-02-05 11:23:08,187 - bootstrapping - INFO - Configuring certificate
2021-02-05 11:23:08,188 - bootstrapping - WARNING - Private key already exists, not overwriting. (Use option --force if necessary)
2021-02-05 11:23:08,188 - bootstrapping - INFO - Configuring wal-e
2021-02-05 11:23:08,188 - bootstrapping - INFO - Configuring bootstrap
2021-02-05 11:23:08,188 - bootstrapping - INFO - Configuring patroni
2021-02-05 11:23:08,197 - bootstrapping - INFO - Writing to file /run/postgres.yml
Traceback (most recent call last):
  File "/scripts/configure_spilo.py", line 1022, in <module>
    main()
  File "/scripts/configure_spilo.py", line 953, in main
    adjust_owner(placeholders, PATRONI_CONFIG_FILE, gid=-1)
  File "/scripts/configure_spilo.py", line 68, in adjust_owner
    os.chown(resource, uid, gid)
PermissionError: [Errno 1] Operation not permitted: '/run/postgres.yml'
seteuid: Operation not permitted
seteuid: Operation not permitted
```

Files inside container:
```shell
postgres@postgres-service-01-0:/run$ ll
total 8
[...]

PR #555, opened 2021-02-14
Fixed issues with arbitrary UIDs

djcass44:
Fixes an issue (see #551) with Spilo running in OpenShift when the UID is a random number picked by OpenShift. This PR reverts the method of determining what uid/gid to use back to checking the `/etc/passwd` file rather than the ownership of the `PGHOME` directory.

Verification that it works:
* Tests all pass
* `docker build -t spilo . && docker run --user=1500080000:0 spilo` is able to start the database with no issues

### DIFF
--- a/postgres-appliance/scripts/configure_spilo.py
+++ b/postgres-appliance/scripts/configure_spilo.py
@@ -12,6 +12,7 @@ import psutil
 import socket
 import subprocess
 import sys
+import pwd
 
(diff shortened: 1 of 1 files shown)


Jan-M:
:+1:

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you @djcass44!

PR #557, opened 2021-02-17
Fix bug with major upgrade after clone

CyberDem0n:
When working on the code removing bg_mon from shared_preload_libraries before executing pg_upgrade a little bug was introduced. Specifically, shared_preload_libraries for the old version was overwritten by the value taken from the new version. As a result, the old cluster was failing to start due to missing (not existing) libraries and upgrade was failing.

This commit is fixing the wrong behavior and improves tests to catch similar issues in the future.


FxKu:
👍

CyberDem0n:
:+1:

PR #572, opened 2021-03-26
Update postgres_backup.sh usage

OlleLarsson:
Minor update to the usage string in the postgres_backup.sh script.
In this [commit](https://github.com/zalando/spilo/commit/84668553b9cda6ba9da199ff3b3fe9fd89cf5aec#diff-12f469d988840b9eb6a93e31e87d9160d606abcdca4e47d6c9224e3df86335f0R14) the way that `DAYS_TO_RETAIN` is set was changed from command line argument to reading from the env var `BACKUP_NUM_TO_RETAIN`.
However, the usage string for the script was not updated to reflect the change, this commit fixes that.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   tests:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-18.04
 
     steps:
     - uses: actions/checkout@v1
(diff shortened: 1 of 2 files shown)


CyberDem0n:
@OlleLarsson good catch, thank you!

CI failure is unrelated to your change, but you can "fix" it by changing `ubuntu-latest` to `ubuntu-18.04` here: https://github.com/zalando/spilo/blob/master/.github/workflows/tests.yaml#L11

OlleLarsson:
Yeah, I updated the gihub workflow as well  :)

sdudoladov:
:+1:

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you @OlleLarsson!

PR #573, opened 2021-03-26
Add backend_type column to postgres_log if it is missing on postgres 13

CyberDem0n:
Due to the bug in the upgrade script postgres_log table wasn't dropped before pg_upgrade. As a result it wasn't recreated with the new structure.
In order to fix it, the post_init.sh running against Postgres 13 script will always try to add a column with NOT EXISTS.

At the same time we simplify the upgrade script. We don't need to drop the postgres_log foreign table before the upgrade.


sdudoladov:
:+1:

CyberDem0n:
:+1:

Issue #562, opened 2021-03-09
seteuid: Operation not permitted but cluster is up and running fine with PG 1.6


neelasha-09:
Hi Team,

We are using  the new PG version 1.6. Below are the images used for OPR and Cluster on OCP environment.
```
OPR :  v1.6.1-23-g2efa8312
Cluster: spilo-13:2.0-p5
```
We see the Cluster is up and running fine, but there are still errors  in cluster logs `seteuid: Operation not permitted`. Logs attached.

[Logs.txt](https://github.com/zalando/spilo/files/6106198/Logs.txt)

Permissions inside cluster:

```sh
postgres@postgres-operator-cluster-1-0:~$ id
uid=1000600000(postgres) gid=0(root) groups=0(root),1000600000
```

Could you please support?

CyberDem0n:
List of processes in the container might help to figure it out.
Please run `ps auxwf` and copy the output here.

neelasha-09:
Please find the output requested.
```sh
postgres@postgres-operator-cluster-1-0:~$ ps auxwf
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
postgres    1390  0.3  0.0   4640   840 pts/0    Ss   07:30   0:00 /bin/sh -c TERM="xterm" /bin/sh
postgres    1397  0.0  0.0   4640   840 pts/0    S    07:30   0:00  \_ /bin/sh
postgres    1399  0.0  0.0  22044  4164 pts/0    S    07:30   0:00      \_ bash
postgres    1417  0.0  0.0  37812  3336 pts/0    R+   07:30   0:00          \_ ps auxwf
postgres       1  0.0  0.0   4396   824 ?        Ss   07:12   0:00 /usr/bin/dumb-init -c --rewrite 1:0 -- /bin/sh /launch.sh
postgres      10  0.0  0.0   4640  1752 ?        S    07:12   0:00 /bin/sh /launch.sh
postgres      32  0.0  0.0   4564   740 ?        S    07:12   0:00  \_ /usr/bin/runsvdir -P /etc/service
postgres      33  0.1  0.0   4412  1288 ?        Ss   07:12   0:01      \_ runsv cron
postgres      34  0.0  0.0   4412   800 ?        Ss   07:12   0:00      \_ runsv pgqd
postgres      38  0.0  0.0 108012  8136 ?        S    07:12   0:00      |   \_ /usr/bin/pgqd /home/postgres/pgq_ticker.ini
[...]

PR #574, opened 2021-03-31
Make it possible to run cron without a root

CyberDem0n:
* Don't set suid bit on `/usr/sbin/cron` (even if it could be used there is no way to safely create the root cronjob, but it also means that the `sys_nice` capability which is set on `chrt` and `renice` could be used).
* Use `LD_PRELOAD` trick for hooking `seteuid()` and `initgroups()` functions when starting cron deamon without a root.
* Enhance scheduling (or not scheduling) the renice cronjob depending on the initial conditions (root, sys_nice, no-new-privileges).
* Disable unwanted cronjobs (`/etc/crontab`, `/etc/cron.{hourly,daily,...}`), they produce "scary" logs when cron is running without a root.

Close: https://github.com/zalando/spilo/issues/562


Jan-M:
:+1:

CyberDem0n:
:+1:

Issue #577, opened 2021-04-16
Propagate WALG_S3_CA_CERT_FILE to /run/etc/wal-e.d/env


stephan2012:
Spilo should propagate `WALG_S3_CA_CERT_FILE` to `/run/etc/wal-e.d/env` (see the [wal-g configuration](https://github.com/wal-g/wal-g#configuration)). This would allow storing backups in on-prem and TLS-enabled [Minio](https://github.com/minio/minio) setups with an internal CA.

When `WALG_S3_CA_CERT_FILE` is not set, `wal-g` hangs. Manually creating`/run/etc/wal-e.d/env/WALG_S3_CA_CERT_FILE` makes backup succeeding.

Maybe I missed a better way of passing a CA certificate?

Spilo image is `registry.opensource.zalan.do/acid/spilo-13:2.0-p6`.

PR #578, opened 2021-04-16
Propagate WALG_S3_CA_CERT_FILE

stephan2012:
Required if the S3 (compatible) service is secured with TLS and uses a
private CA

Closes https://github.com/zalando/spilo/issues/577


ENVIRONMENT.rst
```diff
@@ -42,7 +42,7 @@ Environment Configuration Settings
 - **AWS_ENDPOINT**: (optional) in format 'https://s3.AWS_REGION.amazonaws.com:443', if not specified will be calculated from AWS_REGION
 - **WALE_S3_ENDPOINT**: (optional) in format 'https+path://s3.AWS_REGION.amazonaws.com:443', if not specified will be calculated from AWS_ENDPOINT or AWS_REGION
 - **WALE_S3_PREFIX**: (optional) the full path to the backup location on S3 in the format s3://bucket-name/very/long/path. If not specified Spilo will generate it from WAL_S3_BUCKET.
-- **WALG_DELTA_MAX_STEPS**, **WALG_DELTA_ORIGIN**, **WALG_DOWNLOAD_CONCURRENCY**, **WALG_UPLOAD_CONCURRENCY**, **WALG_UPLOAD_DISK_CONCURRENCY**: (optional) configuration options for wal-g. If at least of the options is specified, wal-g will be used instead of wal-e.
+specified, wal-g will be used instead of wal-e.
```
CyberDem0n, line 45: wal-g is used by default only for restore

ENVIRONMENT.rst
```diff
@@ -83,3 +81,11 @@ Environment Configuration Settings
...
 - **INITDB_LOCALE**: database cluster's default UTF-8 locale (en_US by default)
 - **ENABLE_WAL_PATH_COMPAT**: old Spilo images were generating wal path in the backup store using the following template ``/spilo/{WAL_BUCKET_SCOPE_PREFIX}{SCOPE}{WAL_BUCKET_SCOPE_SUFFIX}/wal/``, while new images adding one additional directory (``{PGVERSION}``) to the end. In order to avoid (unlikely) issues with restoring WALs (from S3/GC/and so on) when switching to ``spilo-13`` please set the ``ENABLE_WAL_PATH_COMPAT=true`` when deploying old cluster with ``spilo-13`` for the first time. After that the environment variable could be removed. Change of the WAL path also mean that backups stored in the old location will not be cleaned up automatically.
+
+wal-g
+-----
+
+- **WALG_AZ_PREFIX**: (optional) the azure prefix to store WAL backups at in the format azure://test-container/walg-folder.
+- **WALG_DELTA_MAX_STEPS**, **WALG_DELTA_ORIGIN**, **WALG_DOWNLOAD_CONCURRENCY**, **WALG_UPLOAD_CONCURRENCY**, **WALG_UPLOAD_DISK_CONCURRENCY**: (optional) configuration options for wal-g. If at least of the options is
```
CyberDem0n, line 89:
```suggestion
- **WALG_DELTA_MAX_STEPS**, **WALG_DELTA_ORIGIN**, **WALG_DOWNLOAD_CONCURRENCY**, **WALG_UPLOAD_CONCURRENCY**, **WALG_UPLOAD_DISK_CONCURRENCY**: (optional) configuration options for wal-g.
```

ENVIRONMENT.rst
stephan2012: Sure, and I'm fine with your changes since I haven written them. Just moved wal-g options into a dedicated section. Or did I miss the point?

ENVIRONMENT.rst
stephan2012: Just reformatting?

ENVIRONMENT.rst
CyberDem0n: I think there is something wrong with formatting.
The `If at least of the options is specified, wal-g will be used instead of wal-e` sentence was split into two and one part was moved to the paragraph about wal-g.

Probably we should remove `specified, wal-g will be used instead of wal-e.` from here completely

ENVIRONMENT.rst
CyberDem0n: No, this is a part of the sentence from above. Now it makes very little sense.

I think we should add the following to the `wal-g` section:
```
The `wal-g` is used by default for Azure and SSH backups and restore.
In case of S3 wal-e is user for backups and wal-g for restore.
```

Also please add two following parameters:
- `USE_WALG_BACKUP`
- `USE_WALG_RESTORE`

They are both "booleans"  and enforce using wal-g instead of wal-e if `true` is specified.

ENVIRONMENT.rst
stephan2012: Got it. Changed in the latest commit.

ENVIRONMENT.rst
stephan2012: Added.

ENVIRONMENT.rst
```diff
@@ -83,3 +80,16 @@ Environment Configuration Settings
...
+wal-g
+-----
+
+`wal-g` is used by default for Azure and SSH backups and restore.
+In case of S3, `wal-e` is used for backups and `wal-g` for restore.
+
+- **USE_WALG_BACKUP**: (optional) Enforce using `wal-g` instead of `wal-e` for backups (Boolean)
+- **USE_WALG_RESTORE**: (optional) Enforce using `wal-g` instead of `wal-e` for restores (Boolean)
+- **WALG_AZ_PREFIX**: (optional) the azure prefix to store WAL backups at in the format azure://test-container/walg-folder.
+- **WALG_DELTA_MAX_STEPS**, **WALG_DELTA_ORIGIN**, **WALG_DOWNLOAD_CONCURRENCY**, **WALG_UPLOAD_CONCURRENCY**, **WALG_UPLOAD_DISK_CONCURRENCY**: (optional) configuration options for wal-g. If at least of the options is specified, wal-g will be used instead of wal-e.
```
CyberDem0n, line 93:
```suggestion
- **WALG_DELTA_MAX_STEPS**, **WALG_DELTA_ORIGIN**, **WALG_DOWNLOAD_CONCURRENCY**, **WALG_UPLOAD_CONCURRENCY**, **WALG_UPLOAD_DISK_CONCURRENCY**: (optional) configuration options for wal-g.
```

I suggest removing it because apparently it is not true.

Jan-M:
:+1:

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you @stephan2012!

PR #580, opened 2021-04-30
Fix clone_with_wale on pg13+ when recovery_time in future

CyberDem0n:
PostgreSQL 13 changed its behavior when the recovery target is unreachable. While old versions were finishing the recovery and
promoting Postgres, starting from 13 Postgres complains "recovery ended before configured recovery target was reached" in logs and exits. As a result, the clone completely fails.

In order to retain the same behavior as with old versions, the `maybe_pg_upgrade.py` will always start point-in-time recovery (even if there is no major upgrade required).
If it failed, for PostgreSQL 13+ it will check a few last log lines and if they contain the error about the recovery target not reached,
it will reset all recovery targets and start the postgres up one more time. The next failure is fatal.

### DIFF
--- a/postgres-appliance/bootstrap/maybe_pg_upgrade.py
+++ b/postgres-appliance/bootstrap/maybe_pg_upgrade.py
@@ -1,14 +1,73 @@
 #!/usr/bin/env python
+import datetime
 import logging
+import os
+import subprocess
 import sys
 
 logger = logging.getLogger(__name__)
 
 
+def tail_postgres_log(weekday):
+    logdir = os.environ.get('PGLOG', '/home/postgres/pgdata/pgroot/pg_log')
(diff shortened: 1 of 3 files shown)


sdudoladov:
:+1:

sdudoladov:
:+1:

CyberDem0n:
:+1:

OlleLarsson:
I haven't run the version of spilo where this pr is included but I was just wondering if this works even if I have disabled csv logging, aka
```
log_destination: "stderr"
logging_collector: "off"
```
?

CyberDem0n:
> if this works even if I have disabled csv logging

No, it won't work. Unfortunately, there is no universal solution to this problem. If you don't rely on CSV logs I can only suggest not set recovery target in the future.

OlleLarsson:
Thanks for the reply, I'll make sure to make a note of this 🙂

PR #581, opened 2021-05-02
Update StatefulSet apiVersion in Kubernetes manifest

ohkinozomu:
- Update deprecated apps/v1beta1 API.
https://kubernetes.io/docs/reference/using-api/deprecation-guide/#statefulset-v116
- Add a simple test for installation on Kubernetes.
- Format affinity JSON.

### DIFF
--- a/.github/workflows/kubernetes-test.yaml
+++ b/.github/workflows/kubernetes-test.yaml
@@ -1,0 +1,22 @@
+name: Kubernetes Installation Test
+
+on: 
+  pull_request:
+    paths:
+    - 'kubernetes/**'
+  push:
+    branches:
+    - master
+    paths:
+    - 'kubernetes/**'
+
+jobs:
+  kubernetes-installation-test:
+    runs-on: ubuntu-20.04
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Create Kind Cluster
+      uses: helm/kind-action@v1.1.0
+    - name: Install Spilo
+      run: kubectl apply -f kubernetes/spilo_kubernetes.yaml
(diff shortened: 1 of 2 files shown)


.github/workflows/kubernetes-test.yaml
```diff
@@ -0,0 +1,22 @@
...
+jobs:
+  kubernetes-installation-test:
+    runs-on: ubuntu-20.04
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Create Kind Cluster
+      uses: helm/kind-action@v1.1.0
+    - name: Install Spilo
+      run: kubectl apply -f kubernetes/spilo_kubernetes.yaml
```
Jan-M, line 22: Should we have some "test" here in the end maybe to see of Postgres is up? e.g. query pods with master label?

CyberDem0n: I don't think this is necessary. We Already exercise Spilo in many different places on K8s.
This manifest is just for people who want to deploy it manually.

ohkinozomu: The main purpose of this test is to make sure that at least "kubectl apply" is successful (deprecated APIs are not used, etc.).
If the manifest file will be maintained in the future, it is good to have the certainty of apply guaranteed.

I used kind because that's the most reliable way, but I agree that it can be misleading that this is an spilo E2E test (my naming may be inappropriate).
If I should limit the scope to detecting deprecated APIs, other tools may be better ([kubeent](https://github.com/doitintl/kube-no-trouble)?).

sdudoladov:
:+1:

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you @ohkinozomu!

Issue #571, opened 2021-03-21
Cron based backups failing when using non-default postgres port


KarlHerler:
When running postgres on a non default port the `PGPORT` environment variable is set to the given port. The backup script relies on it being correct, even when connecting over unix sockets, as the backup script `/scripts/postgres_backup.sh` does. The environment setup for cron lacks the environmental variable.

When running with another port than the default the cron job will result in this being printed on `stderr`:

```
psql: error: could not connect to server: No such file or directory
	Is the server running locally and accepting
	connections on Unix domain socket "/var/run/postgresql/.s.PGSQL.5432"?
```

However when running it manually it works excellently. This is due to the `PGPORT` being set in the default users but is missing in cron.

An easy fix to this is to a file called `PGPORT` with the port used to `/run/etc/wal-e.d/env`. However this folder is not persistently stored so it need to be repopulated after restarts. 

I would like to implement this fix properly, however I am unsure where exactly the content of  `/run/etc/wal-e.d/env` is generated.

CyberDem0n:
Most likely adding it [here](https://github.com/zalando/spilo/blob/master/postgres-appliance/scripts/configure_spilo.py#L708-L710) should be enough.

PR #582, opened 2021-05-12
Add support for WAL-E/WAL-G with custom PGPORT

KarlHerler:
Fixes issue #571.

The comment: https://github.com/zalando/spilo/issues/571#issuecomment-806790634 was not sufficient as `PGPORT` must also be placed in `write_envidir_names` or as implemented here.

Note that the current implementation adds  the `PGPORT` file to the WAL-E/WAL-G configuration files indiscriminately, even if it isn't necessary for default port use cases. This does not cause any problems but it is redundant.


sdudoladov:
:+1:

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you @KarlHerler!

PR #584, opened 2021-05-19
Fix issue with PITR

CyberDem0n:
if the PITR performed by the maybe_pg_upgrade.py script and not followed by the major upgrade the archive_mode was set to off.


FxKu:
:+1:

CyberDem0n:
:+1:

PR #607, opened 2021-08-10
Add timescaledb 2.4.0

CyberDem0n:
* build all possible latest major versions of timescaledb for every major version of Postgres (sometimes they are required for major upgrade)
* change post_init.sh script to not rely on TIMESCALEDB* env variables for determining whether the timescaledb extension should be updated


Jan-M:
:+1:

CyberDem0n:
:+1:

PR #624, opened 2021-09-08
Document SSL_CERTIFICATE and SSL_PRIVATE_KEY and add SSL_CA and SSL_C…

sasadangelo:
Currently, Spilo has two variables not documented:
* SSL_CERTIFICATE
* SSL_PRIVATE_KEY

you can use it to pass a custom certificate with a private key.
The problem is that there is no variable to pass the content of the CA certificate and let Spilo create it in SSL_CA_FILE path. For this reason, I added the following variable:
* SSL_CA

Alexander Kukushkin asked me on the slack channel to add the SSL_CRL variable as well.
Please review the PR and let me know for any issue.

### DIFF
--- a/postgres-appliance/scripts/configure_spilo.py
+++ b/postgres-appliance/scripts/configure_spilo.py
@@ -112,7 +112,6 @@ def write_certificates(environment, overwrite):
         p = subprocess.Popen(openssl_cmd, shell=False, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
         output, _ = p.communicate()
         logging.debug(output)
-        logging.info('Generating ssl certificate')
     if 'SSL_CA' in environment:
         logging.info('Generating ssl ca certificate')
         write_file(environment['SSL_CA'], environment['SSL_CA_FILE'], overwrite)
--- a/postgres-appliance/scripts/configure_spilo.py
+++ b/postgres-appliance/scripts/configure_spilo.py
@@ -91,6 +91,16 @@ def write_certificates(environment, overwrite):
         logging.info('Writing custom ssl certificate')
         for k in ssl_keys:
             write_file(environment[k], environment[k + '_FILE'], overwrite)
+        if 'SSL_CA' in environment:
+            logging.info('Generating ssl ca certificate')
+            write_file(environment['SSL_CA'], environment['SSL_CA_FILE'], overwrite)
+        else:
+            logging.info('No CA certificate to generate')
+        if 'SSL_CRL' in environment:
+            logging.info('Generating ssl crl certificate')
+            write_file(environment['SSL_CRL'], environment['SSL_CRL_FILE'], overwrite)
+        else:
+            logging.info('No CRL certificate to generate')
     else:
         if os.path.exists(environment['SSL_PRIVATE_KEY_FILE']) and not overwrite:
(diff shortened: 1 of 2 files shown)


postgres-appliance/scripts/configure_spilo.py
```diff
@@ -107,10 +108,21 @@ def write_certificates(environment, overwrite):
             '-out',
             environment['SSL_CERTIFICATE_FILE'],
         ]
-        logging.info('Generating ssl certificate')
+        logging.info('Autogenerating ssl certificate')
         p = subprocess.Popen(openssl_cmd, shell=False, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
         output, _ = p.communicate()
         logging.debug(output)
+        logging.info('Generating ssl certificate')
```
CyberDem0n, line 115: Redundant message, duplicates https://github.com/zalando/spilo/pull/624/files#diff-e4dcf964214d58d733ae9874d4a2ab4b8e9eb99213307e1666af5dad21a326bbR111
```suggestion
```

postgres-appliance/scripts/configure_spilo.py
```diff
@@ -88,6 +88,7 @@ def write_certificates(environment, overwrite):
 
     ssl_keys = ['SSL_CERTIFICATE', 'SSL_PRIVATE_KEY']
     if set(ssl_keys) <= set(environment):
+        logging.info('Generating custom ssl certificate')
```
CyberDem0n, line 91: We are not generating anything here, only writing to files.

postgres-appliance/scripts/configure_spilo.py
```diff
@@ -107,10 +108,21 @@ def write_certificates(environment, overwrite):
...
+    if 'SSL_CA' in environment:
+        logging.info('Generating ssl ca certificate')
+        write_file(environment['SSL_CA'], environment['SSL_CA_FILE'], overwrite)
+    else:
+        logging.info('No ca certificate to generate')
+    if 'SSL_CRL' in environment:
+        logging.info('Generating ssl crl certificate')
+        write_file(environment['SSL_CRL'], environment['SSL_CRL_FILE'], overwrite)
+    else:
+        logging.info('No crl certificate to generate')
```
CyberDem0n, line 125: This block should not be executed if is self-signed certificates are generated.
The better place for it would be here: https://github.com/zalando/spilo/pull/624/files#diff-e4dcf964214d58d733ae9874d4a2ab4b8e9eb99213307e1666af5dad21a326bbR92-R93

Also, default values of `SSL_CA_FILE` and `SSL_CRL_FILE` are empty strings.
If there is a content for them they should be set to `os.path.join(placeholders['RW_DIR'], 'certs', 'ca.crt'))` and `os.path.join(placeholders['RW_DIR'], 'certs', 'server.crl'))` respectively.

postgres-appliance/scripts/configure_spilo.py
sasadangelo: Is it ok for you if I fix this with "Writing custom SSL certificate"

postgres-appliance/scripts/configure_spilo.py
sasadangelo: Is it OK for you if I remove it?

sasadangelo:
I close this PR to submit a new one.

PR #625, opened 2021-09-09
Document SSL_CERTIFICATE and SSL_PRIVATE_KEY and add SSL_CA and SSL_C…

sasadangelo:
Currently, Spilo has two variables not documented:

SSL_CERTIFICATE
SSL_PRIVATE_KEY
you can use it to pass a custom certificate with a private key.
The problem is that there is no variable to pass the content of the CA certificate and let Spilo create it in the SSL_CA_FILE path. For this reason, I added the following variable:

SSL_CA
Alexander Kukushkin asked me on the slack channel to add the SSL_CRL variable as well.
Please review the PR and let me know of any issues.

### DIFF
--- a/postgres-appliance/scripts/configure_spilo.py
+++ b/postgres-appliance/scripts/configure_spilo.py
@@ -88,8 +88,19 @@ def write_certificates(environment, overwrite):
 
     ssl_keys = ['SSL_CERTIFICATE', 'SSL_PRIVATE_KEY']
     if set(ssl_keys) <= set(environment):
+        logging.info('Writing custom ssl certificate')
         for k in ssl_keys:
             write_file(environment[k], environment[k + '_FILE'], overwrite)
+        if 'SSL_CA' in environment:
+            logging.info('Writing ssl ca certificate')
+            write_file(environment['SSL_CA'], environment['SSL_CA_FILE'], overwrite)
+        else:
+            logging.info('No ca certificate to write')
+        if 'SSL_CRL' in environment:
+            logging.info('Writing ssl certificate revocation list')
+            write_file(environment['SSL_CRL'], environment['SSL_CRL_FILE'], overwrite)
+        else:
+            logging.info('No certificate revocation list to write')
     else:
         if os.path.exists(environment['SSL_PRIVATE_KEY_FILE']) and not overwrite:
             logging.warning('Private key already exists, not overwriting. (Use option --force if necessary)')
@@ -107,7 +118,7 @@ def write_certificates(environment, overwrite):
             '-out',
             environment['SSL_CERTIFICATE_FILE'],
         ]
-        logging.info('Generating ssl certificate')
+        logging.info('Generating ssl self-signed certificate')
         p = subprocess.Popen(openssl_cmd, shell=False, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
         output, _ = p.communicate()
         logging.debug(output)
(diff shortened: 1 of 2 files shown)


postgres-appliance/scripts/configure_spilo.py
```diff
@@ -88,8 +88,23 @@ def write_certificates(environment, overwrite):
 
     ssl_keys = ['SSL_CERTIFICATE', 'SSL_PRIVATE_KEY']
     if set(ssl_keys) <= set(environment):
+        logging.info('Writing custom ssl certificate')
         for k in ssl_keys:
             write_file(environment[k], environment[k + '_FILE'], overwrite)
+        if 'SSL_CA' in environment:
+            if environment['SSL_CA_FILE'] == '':
+                environment['SSL_CA_FILE'] = os.path.join(environment['RW_DIR'], 'certs', 'ca.crt')
```
CyberDem0n, line 96: Unfortunately, the `write_certificates()` function is called after the `TEMPLATE` is evaluated and hence if `SSL_CA` is set, but `SSL_CA_FILE` isn't set, the file will be written, but postgres.yml will lack `postgresql.parameters.ssl_ca_file`.

The same applies to `SSL_CRL_FILE`.

Such checks should be moved to the `get_placeholders()` function.

sasadangelo: ok

postgres-appliance/scripts/configure_spilo.py
```diff
@@ -612,6 +623,15 @@ def get_placeholders(provider):
 
     placeholders['BGMON_LISTEN_IP'] = get_listen_ip()
 
+    if 'SSL_CA' in placeholders and placeholders['SSL_CA_FILE'] == '':
+        placeholders['SSL_CA_FILE'] = os.path.join(placeholders['RW_DIR'], 'certs', 'ca.crt')
+    if not 'SSL_CA' in placeholders and placeholders['SSL_CA_FILE'] != '':
+        placeholders['SSL_CA_FILE'] = ''
```
CyberDem0n, line 629: This one should be removed. It is quite common that people just mount certificates and provide their location in the `*_FILE` env var.

sasadangelo: ok as you prefer, it was just to solve a possible inconsistency in input parameters

sasadangelo: fix requested done

postgres-appliance/scripts/configure_spilo.py
```diff
@@ -88,8 +88,19 @@ def write_certificates(environment, overwrite):
...
+        if 'SSL_CA' in environment:
+            logging.info('Writing ssl ca certificate')
+            write_file(environment['SSL_CA'], environment['SSL_CA_FILE'], overwrite)
+        else:
+            logging.info('No ca certificate to write')
+        if 'SSL_CRL' in environment:
+            logging.info('Writing ssl certificate revocation list')
+            write_file(environment['SSL_CRL'], environment['SSL_CRL_FILE'], overwrite)
+        else:
+            logging.info('No crl to write')
```
Jan-M, line 103: lets also write 'certificate revocation list' here

sasadangelo: Hi @Jan-M, fix done. Check commit, please.

Jan-M:
:+1:

CyberDem0n:
👍

CyberDem0n:
Merged, thank you @sasadangelo!

PR #633, opened 2021-09-21
Remove PostgreSQL v9.5, which is EOL since Feb 2021

CyberDem0n:
- remove a couple of extensions installed with pgxn (the build fails on
v11+ anyway)
- update tests


sdudoladov:
:+1:

CyberDem0n:
:+1:

PR #634, opened 2021-09-21
Add PGroonga extension

Dador:
PGroonga is quite popular extension for full text search in Japanese, Chinese and other asian languages. This PR adds it to spilo image.

From [official website](https://pgroonga.github.io/):
> PostgreSQL supports full text search against languages that use only alphabet and digit. It means that PostgreSQL doesn't support full text search against Japanese, Chinese and so on. You can use super fast full text search feature against all languages by installing PGroonga into your PostgreSQL!

[PGroonga's GitHub](https://github.com/pgroonga/pgroonga)

### DIFF
--- a/postgres-appliance/Dockerfile
+++ b/postgres-appliance/Dockerfile
@@ -53,6 +53,17 @@ RUN export DEBIAN_FRONTEND=noninteractive \
     done \
     && curl -s -o - https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add - \
 \
+    # Add pgroonga repositories
+    # Libs + for version 9.6
+    && for t in deb deb-src; do \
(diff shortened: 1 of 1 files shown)


Dador:
There was similar PR (#493) with other extension from external source and also there is several extensions from external sources in `Dockerfile` itself, so I thought you accept new extensions if they have some popularity.

PGroonga's packages are build for official PostgreSQL packages from `apt.postgresql.org` ([source](https://pgroonga.github.io/install/ubuntu.html#install-for-official-postgresql)) so there shouldn't be any incompatibilities. PGroonga itself is quite stable and not new project, also it's maintained by members @kou and @komainu8 of original project ([Groonga](https://github.com/groonga/groonga)).

But I understand your concern about future support. After all it's just a suggestion and of course decision is up to you.

kou:
I'm not sure the context but feel free to ask us about PGroonga.

PR #640, opened 2021-09-28
Define RESTAPI_CONNECT_ADDRESS environment variable

sasadangelo:
Spilo by default uses IP for the property ```restapi.connect_address```. This is a problem when Patroni REST API is configured in SSL with client authentication because hostname validation is performed. Due to the dynamic nature of Pod, it's impossible to create SSL certificates with the right IP into SAN.
This PR solves the problem with the introduction of the property: REST_API_CONNECT_ADDRESS.
You can configure it, for example, like this:
```
        - name: RESTAPI_CONNECT_ADDRESS
          value: "$(POD_NAME).<service name>"
``` 

so that it uses the <pod name>.<service name> stored in CoreDNS.

### DIFF
--- a/postgres-appliance/scripts/configure_spilo.py
+++ b/postgres-appliance/scripts/configure_spilo.py
@@ -264,7 +264,7 @@ bootstrap:
 scope: &scope '{{SCOPE}}'
 restapi:
   listen: ':{{APIPORT}}'
-  connect_address: {{REST_API_CONNECT_ADDRESS}}:{{APIPORT}}
+  connect_address: {{RESTAPI_CONNECT_ADDRESS}}:{{APIPORT}}
   {{#SSL_RESTAPI_CA_FILE}}
   cafile: {{SSL_RESTAPI_CA_FILE}}
   {{/SSL_RESTAPI_CA_FILE}}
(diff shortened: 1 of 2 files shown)


postgres-appliance/scripts/configure_spilo.py
```diff
@@ -264,7 +264,7 @@ def deep_update(a, b):
 scope: &scope '{{SCOPE}}'
 restapi:
   listen: ':{{APIPORT}}'
-  connect_address: {{instance_data.ip}}:{{APIPORT}}
+  connect_address: {{REST_API_CONNECT_ADDRESS}}:{{APIPORT}}
```
CyberDem0n, line 267: I think it is better to call it `RESTAPI_CONNECT_ADDRESS` so it looks similar to https://patroni.readthedocs.io/en/latest/ENVIRONMENT.html#rest-api

sasadangelo: OK done. I added also a row in ENVIRONMENT.rst for documentation.
Please check it.

FxKu:
:+1:

CyberDem0n:
:+1:

PR #641, opened 2021-09-30
Update basebackup.sh

ser-sc:
Fix basebackup method for replica


CyberDem0n:
If pg_basebackup is streaming and the streaming subprocess fails, the whole basebackup is failing.
In order to mitigate the problem, we start pg_receivewal in parallel to pg_basebackup, hence --wal-method is deliberately set to none.

PR #669, opened 2021-12-02
docs: WAL_S3_BUCKET is not s3:// path

huww98:
We will add "s3://" to the *_PREFIX \[1], so it should not appear in WAL_S3_BUCKET.

Besides, "Spilo will add /spilo/scope/wal to that path" is not true now (we added PGVERSION). So just remove it.

\[1]: https://github.com/zalando/spilo/blob/9f37ab473caea1e4420d8f4da8773a4938d9310b/postgres-appliance/scripts/configure_spilo.py#L830


FxKu:
:+1:

CyberDem0n:
:+1:

PR #670, opened 2021-12-02
set AWS_S3_FORCE_PATH_STYLE accordingly

huww98:
WAL-G will not work if we use our own endpoint that only supports path convention but AWS_S3_FORCE_PATH_STYLE is not set. This patch ensures that WAL-G and WAL-E will use the same convention.

Also add support to WALE_S3_ENDPOINT with convention other than path.

Signed-off-by: 胡玮文 <huww98@outlook.com>


FxKu:
:+1:

CyberDem0n:
:+1:

PR #677, opened 2021-12-31
Draft: Replace cron by supercronic to support disallowing privilege escalation

haslersn:
Related to: https://github.com/zalando/postgres-operator/issues/1730

Note that this is currently not tested, thus work-in-progress. However, I'd like to get feedback about this approach.

### DIFF
--- a/postgres-appliance/Dockerfile
+++ b/postgres-appliance/Dockerfile
@@ -59,12 +59,14 @@ ARG ADDITIONAL_LOCALES=
 RUN export DEBIAN_FRONTEND=noninteractive \
     && echo 'APT::Install-Recommends "0";\nAPT::Install-Suggests "0";' > /etc/apt/apt.conf.d/01norecommend \
     && apt-get update \
-    && apt-get install -y curl ca-certificates less locales jq vim-tiny gnupg1 runit dumb-init libcap2-bin rsync sysstat \
+    && apt-get install -y curl ca-certificates less locales jq vim-tiny gnupg1 cron runit dumb-init libcap2-bin rsync sysstat \
     && ln -s chpst /usr/bin/envdir \
     # Make it possible to use the following utilities without root (if container runs without "no-new-privileges:true")
     && setcap 'cap_sys_nice+ep' /usr/bin/chrt \
     && setcap 'cap_sys_nice+ep' /usr/bin/renice \
-\
+    # Disable unwanted cron jobs
+    && rm -fr /etc/cron.??* \
+    && truncate --size 0 /etc/crontab \
     # Install supercronic
     && curl -L https://github.com/aptible/supercronic/releases/download/v0.1.12/supercronic-linux-amd64 > /usr/bin/supercronic \
     && chmod +x /usr/bin/supercronic \
@@ -113,7 +115,7 @@ RUN export DEBIAN_FRONTEND=noninteractive \
             /usr/share/locale/??_?? \
     && find /var/log -type f -exec truncate --size 0 {} \;
 
-COPY dependencies/src /builddeps/
+COPY cron_unprivileged.c dependencies/src /builddeps/
 COPY --from=dependencies-builder /builddeps/*.deb /builddeps/
 
 ARG PGVERSION
@@ -296,6 +298,9 @@ RUN export DEBIAN_FRONTEND=noninteractive \
             done; \
         done; \
     done \
+\
+    # make it possible for cron to work without root
+    && gcc -s -shared -fPIC -o /usr/local/lib/cron_unprivileged.so cron_unprivileged.c \
 \
     # Remove unnecessary packages
     && apt-get purge -y ${BUILD_PACKAGES} postgresql postgresql-all postgresql-server-dev-* libpq-dev=* libmagic1 bsdmainutils \
@@ -506,8 +511,9 @@ COPY pgq_ticker.ini $PGHOME/
 
 RUN sed -i "s|/var/lib/postgresql.*|$PGHOME:/bin/bash|" /etc/passwd \
         && chown -R postgres:postgres $PGHOME $RW_DIR \
-        && rm -fr /var/tmp \
+        && rm -fr /var/spool/cron /var/tmp \
         && mkdir -p /var/spool \
+        && ln -s $RW_DIR/cron /var/spool/cron \
         && ln -s $RW_DIR/tmp /var/tmp \
         && ln -snf $RW_DIR/service /etc/service \
         && ln -s $RW_DIR/pam.d-postgresql /etc/pam.d/postgresql \
--- a/postgres-appliance/Dockerfile
+++ b/postgres-appliance/Dockerfile
@@ -59,14 +59,15 @@ ARG ADDITIONAL_LOCALES=
 RUN export DEBIAN_FRONTEND=noninteractive \
     && echo 'APT::Install-Recommends "0";\nAPT::Install-Suggests "0";' > /etc/apt/apt.conf.d/01norecommend \
     && apt-get update \
-    && apt-get install -y curl ca-certificates less locales jq vim-tiny gnupg1 cron runit dumb-init libcap2-bin rsync sysstat \
+    && apt-get install -y curl ca-certificates less locales jq vim-tiny gnupg1 runit dumb-init libcap2-bin rsync sysstat \
     && ln -s chpst /usr/bin/envdir \
     # Make it possible to use the following utilities without root (if container runs without "no-new-privileges:true")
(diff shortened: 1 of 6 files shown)


CyberDem0n:
Have you actually checked that it doesn't work without root?
https://github.com/zalando/spilo/commit/c91248e26e2ea910304d04a3acbeda1e965e2e42#diff-dd039299d1b892e68ab0ea5e22d533814e1c349a290b5cc46b9d823578f6b741 solved it about half a year ago

haslersn:
The upstream works without root, however, it doesn't work without privilege escalation (Kubernetes `allowPrivilegeEscalation: false` or Docker `--security-opt=no-new-privileges:true`). This PR is to support disallowing privilege escalation.

CyberDem0n:
> The upstream works without root, however, it doesn't work without privilege escalation

Indeed :(

The `cron` daemon is running, but it fails to start the job due to some errors raised from pam.
Specifically, pam complains about the inability to read `/etc/shadow`.
Since there is no secret information in it, it is safe to give the read permission on this file to all. It makes cron fully functional.
```diff
diff --git a/postgres-appliance/Dockerfile b/postgres-appliance/Dockerfile
index ded4343..8db483d 100644
--- a/postgres-appliance/Dockerfile
+++ b/postgres-appliance/Dockerfile
@@ -534,6 +534,7 @@ RUN sed -i "s|/var/lib/postgresql.*|$PGHOME:/bin/bash|" /etc/passwd \
         # Allow users in the root group to access the following files and dirs
         && if [ "$COMPRESS" != "true" ]; then \
            chmod 664 /etc/passwd \
+           && chmod o+r /etc/shadow \
            && chgrp -R 0 $PGHOME $RW_DIR \
            && chmod -R g=u $PGHOME $RW_DIR \
            && usermod -a -G root postgres; \
```

```bash
$ docker run -u 1234567 --security-opt=no-new-privileges:true -e CRONTAB='["* * * * * date > /tmp/cron.log"]' --rm -ti spilo 
2022-01-01 09:41:27,674 - bootstrapping - INFO - Figuring out my environment (Google? AWS? Openstack? Local?)
2022-01-01 09:41:29,679 - bootstrapping - INFO - Could not connect to 169.254.169.254, assuming local Docker setup
2022-01-01 09:41:29,686 - bootstrapping - INFO - No meta-data available for this provider
2022-01-01 09:41:29,688 - bootstrapping - INFO - Looks like your running local
2022-01-01 09:41:29,764 - bootstrapping - INFO - Configuring log
2022-01-01 09:41:29,764 - bootstrapping - INFO - Configuring pgbouncer
2022-01-01 09:41:29,764 - bootstrapping - INFO - No PGBOUNCER_CONFIGURATION was specified, skipping
2022-01-01 09:41:29,764 - bootstrapping - INFO - Configuring wal-e
2022-01-01 09:41:29,764 - bootstrapping - INFO - Configuring certificate
2022-01-01 09:41:29,764 - bootstrapping - INFO - Generating ssl self-signed certificate
2022-01-01 09:41:29,810 - bootstrapping - INFO - Configuring standby-cluster
2022-01-01 09:41:29,810 - bootstrapping - INFO - Configuring pam-oauth2
2022-01-01 09:41:29,810 - bootstrapping - INFO - No PAM_OAUTH2 configuration was specified, skipping
2022-01-01 09:41:29,810 - bootstrapping - INFO - Configuring patroni
2022-01-01 09:41:29,821 - bootstrapping - INFO - Writing to file /run/postgres.yml
2022-01-01 09:41:29,822 - bootstrapping - INFO - Configuring pgqd
2022-01-01 09:41:29,822 - bootstrapping - INFO - Configuring crontab
2022-01-01 09:41:29,822 - bootstrapping - INFO - Skipping creation of renice cron job due to lack of SYS_NICE capability
2022-01-01 09:41:29,832 - bootstrapping - INFO - Configuring bootstrap
...
```

```bash
$ docker exec -ti peaceful_brown bash
postgres@405792b6a5b6:~$ crontab -l
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/lib/postgresql/14/bin
* * * * * date > /tmp/cron.log
postgres@405792b6a5b6:~$ ls -al /tmp/cron.log 
-rw-r--r-- 1 postgres root 29 Jan  1 09:43 /tmp/cron.log
```

I will keep insisting on using the `cron` instead of `Supercronic` because the last one doesn't allow changing users, but we rely on this feaure.

haslersn:
> I will keep insisting on using the `cron` instead of `Supercronic` because the last one doesn't allow changing users

What exactly do you mean? Currently spilo only supports running jobs as users `root` or `postgres`. If your job runs as `root`, the it can change users, no matter if it's started by cron or supercronic. Or did you mean something else?

haslersn:
Don't get me wrong. Even if the supercronic feature is wanted, then it should probably at first be opt-in via an environment variable or the like.

haslersn:
I added a commit that makes supercronic opt-in via the environment variable `CRON_DRIVER=supercronic`. Omitting it is equivalent to `CRON_DRIVER=cron`.

Furthermore, I made supercronic's environment variables backwards-compatible with cron. This way, supercronic should be a drop-in replacement.

Finally, I did some basic tests to verify that `CRON_DRIVER=cron` and `CRON_DRIVER=supercronic` both work and deliver the same environment. In particular, the following four runs produce the same `/tmp/cron.log` file:

```bash
cd postgres-appliance
docker build . -t spilo-supercronic
docker run --user postgres \
    -e CRONTAB='["* * * * * env > /tmp/cron.log"]' spilo-supercronic
docker run --user root --security-opt no-new-privileges \
    -e CRONTAB='["* * * * * env > /tmp/cron.log"]' spilo-supercronic
docker run -e CRON_DRIVER=supercronic --user postgres --security-opt no-new-privileges \
    -e CRONTAB='["* * * * * env > /tmp/cron.log"]' spilo-supercronic
docker run -e CRON_DRIVER=supercronic --user root --security-opt no-new-privileges \
    -e CRONTAB='["* * * * * env > /tmp/cron.log"]' spilo-supercronic
```

CyberDem0n:
> Currently spilo only supports running jobs as users root or postgres

Wrong, spilo currently does both. There is one job that is being scheduled to run as the root, while for example the postgres_backup.sh and test_reload_ssl.sh are always started from the postgres user even is cron running from root.

> Don't get me wrong.

Don't get me wrong either. Fixing a little issue with the current cron implementation requires adding only one line to the Dockerfile, while you suggest to add/change ~50 new lines.
I don't see a value in having an alternative cron in the image, because:
1. It increases the image size by yet another 14MB and provides no real improvement compared to the normal cron
2. It further increases the complexity of already complex enough configure_spilo.py
3. On top of that, it also increases maintenance costs.

haslersn:
> Wrong, spilo currently does both. There is one job that is being scheduled to run as the root, while for example the postgres_backup.sh and test_reload_ssl.sh are always started from the postgres user even is cron running from root.

That's exactly what I meant. This PR does the same with supercronic: The renice job is running in a supercronic instance run as root, while the other jobs are running in a supercronic instance running as user postgres (even if the container is started as root).

In my opinion, supercronic provides the following advantages over normal cron:

- No need for the `cron_unprivileged.so` hack which might break in a cron update.
- Statically linked, so it's guaranteed to still work when one upgrades the base image and/or changes the distribution at some point.
- Prometheus metrics: Supercronic can be told to expose prometheus metrics by adding `-prometheus-listen-address`.
- Improved logging. Here's an example from the container's stdout:
  ```
  time="2022-01-02T21:05:00Z" level=info msg=starting iteration=5 job.command="bash /scripts/renice.sh" job.position=0 job.schedule="*/5 * * * *"
  time="2022-01-02T21:05:00Z" level=info msg=starting iteration=28 job.command="env > /tmp/cron.log" job.position=1 job.schedule="* * * * *"
  time="2022-01-02T21:05:00Z" level=info msg="job succeeded" iteration=28 job.command="env > /tmp/cron.log" job.position=1 job.schedule="* * * * *"
  time="2022-01-02T21:05:00Z" level=info msg="job succeeded" iteration=5 job.command="bash /scripts/renice.sh" job.position=0 job.schedule="*/5 * * * *"
  ```

> while you suggest to add/change ~50 new lines.

That's only the case until it's verified to work as a drop-in. Then legacy cron support can be dropped and we're at the same number of lines as before. (Actually even slightly below it, due to some lines that can be dropped from the Dockerfile.)

> 2. It further increases the complexity of already complex enough configure_spilo.py
> 3. On top of that, it also increases maintenance costs.

Both is only the case until legacy cron support can be dropped.

However, I agree that the increased image size is a serious problem.

PR #679, opened 2022-01-03
Various fixes

CyberDem0n:
1. Ensure files copied from git repo are not world-writable. The problem seems to be very specific to Zalando CPD, that possibly does git clone with umask=0000
2. Read-only rootfs: create symlinks only to corresponding runit directories instead of individual `run` and `finish` files.
3. Fix bug with the cron daemon running from the postgres user. While the daemon was running since https://github.com/zalando/spilo/commit/c91248e26e2ea910304d04a3acbeda1e965e2e42#diff-dd039299d1b892e68ab0ea5e22d533814e1c349a290b5cc46b9d823578f6b741, individual jobs were failing to start because
PAM was failing to read /etc/shadow.
4.  Run "arm64" build on the "large" instance. Having more CPU cores available significantly speeds up the build.

### DIFF
--- a/delivery.yaml
+++ b/delivery.yaml
@@ -11,6 +11,7 @@ pipeline:
   commands:
   - desc: Build spilo docker image
     cmd: |
+      chmod -R o-w postgres-appliance
       cd postgres-appliance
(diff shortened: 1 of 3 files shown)


sdudoladov:
:+1:

CyberDem0n:
:+1:

PR #686, opened 2022-01-10
Implement non-exclusive backups with wal-e

CyberDem0n:
It makes it redundant to remove the backup_label file from the configure_spilo.py


FxKu:
:+1:

CyberDem0n:
:+1:

PR #689, opened 2022-01-17
Install decoderbufs from pgdg package

CyberDem0n:
and add a few more interesting extensions:
- pgq_node
- hll
- dirtyread
- extra-window-functions


FxKu:
:+1:

CyberDem0n:
:+1:

PR #691, opened 2022-01-25
Add retry to launch.sh for configure_spilo.py

rocket357:
Add a small delay and single retry for /scripts/configure_spilo.py from /launch.sh.

It is possible that a large cluster with iptables-enforced network policies could have an ever so brief delay in enforcing network policies.  In my testing it was possible to trigger in a ~150 node cluster with sufficient regularity to cause additional admin work.  Having a single retry after a few seconds alleviates the issue so we don't have to destroy/recreate the pod (and if it still fails after the retry, we'll destroy/recreate the pod anyways, so the only cost is a brief delay upon initial failure).

This could potentially be improved with exponential backoff retries, but I've opted for the simplest approach at this point in time.  I am, however, open to any and all feedback on the PR.

Related to: https://github.com/zalando/spilo/issues/511

### DIFF
--- a/postgres-appliance/launch.sh
+++ b/postgres-appliance/launch.sh
@@ -46,6 +46,14 @@ elif python3 /scripts/configure_spilo.py all; then
     else
         $CMD &
     fi
+else ## /scripts/configure_spilo failed.  Try one more time after a brief pause.
+    sleep 2
+    CMD="/scripts/patroni_wait.sh -t 3600 -- envdir $WALE_ENV_DIR /scripts/postgres_backup.sh $PGDATA"
+    if [ "$(id -u)" = "0" ]; then
+        su postgres -c "PATH=$PATH $CMD" &
+    else
+        $CMD &
+    fi
 fi
 
 sv_stop() {
--- a/postgres-appliance/launch.sh
+++ b/postgres-appliance/launch.sh
@@ -47,6 +47,7 @@ elif python3 /scripts/configure_spilo.py all; then
         $CMD &
     fi
 else ## /scripts/configure_spilo failed.  Try one more time after a brief pause.
+    echo "/launch.sh: /scripts/configure_spilo.py failed.  Retrying after a brief delay..."
     sleep 2
     CMD="/scripts/patroni_wait.sh -t 3600 -- envdir $WALE_ENV_DIR /scripts/postgres_backup.sh $PGDATA"
     if [ "$(id -u)" = "0" ]; then


rocket357:
Additional information:

namespaces affected have a default deny policy, which appears to be enforced at all times, so bringing up the "allow" policies (one of which is to metadata services) causes the initial configure_spilo.py failure.

postgres-appliance/launch.sh
```diff
@@ -46,6 +46,15 @@ elif python3 /scripts/configure_spilo.py all; then
...
     fi
+else ## /scripts/configure_spilo failed.  Try one more time after a brief pause.
+    echo "/launch.sh: /scripts/configure_spilo.py failed.  Retrying after a brief delay..."
+    sleep 2
+    CMD="/scripts/patroni_wait.sh -t 3600 -- envdir $WALE_ENV_DIR /scripts/postgres_backup.sh $PGDATA"
+    if [ "$(id -u)" = "0" ]; then
+        su postgres -c "PATH=$PATH $CMD" &
+    else
+        $CMD &
+    fi
```
CyberDem0n, line 57: I think you don't understand what you are doing.

Non-zere exit code in configure_spilo.py doesn't always mean that it failed:
https://github.com/zalando/spilo/blob/ab4ddee8eae561373eaa272b5a82d12b6621f8fc/postgres-appliance/scripts/configure_spilo.py#L1110-L1111

If retry should happen, you better do it in the configure_spilo.py itself.

But, if we think a bit harder and analyze the problem, it becomes clear that we are dealing with a classical race condition. Normally such kinds of problems are solved with the help of init containers and spilo shouldn't even start until the full initialization is completed.

rocket357: I did miss the fact that the exit code is being abused in that manner.  That does indeed complicate matters.

I'll investigate the init container approach you've suggested.

rocket357:
My init container is a busybox image that does the following:

command: [ "sh", "-c", "wget http://169.254.169.254/ || sleep 5" ]

wget does retry, so there are sometimes additional delays, but there are times when wget connects immediately and the init container terminates, then the postgres container comes up and fails to connect to metadata.  We've noticed upwards of 1.7 seconds of delay in applying network policies, but it only seems to occur on Ubuntu 20.04 (Debian is considerably faster, always less than 1 second and typically within 500 ms).  This suggests the init container and postgres container network policies do not apply simultaneously.

I'm going to open a new PR with retry implemented in configure_spilo.py.

PR #696, opened 2022-02-10
Make sure the initialize key with the wrong sysid is removed

CyberDem0n:
As a part of major upgrade we remove the initialize key from DCS because the upgraded cluster always gets the new system identifier.
But, there is a chance that Patroni could restore a previous value before it was restarted. Therefore, the upgrade script removes the initialize key the second time right after Patroni was restarted. In most cases (and for the most DCS) it worked fine. The only exception is K8s (and maybe Etcd3), because the delete happens based on the cached state, which is updated asynchronously from watch events.

We can't completely eliminate the race condition, but at least we can significantly reduce the chance of getting the cluster to a broken state by doing yet another attempt of removing the initialize key before resume.

Last, but not least, we also reduce chances of using the stale cache by tuning tcp timeouts and keepalives.


CyberDem0n:
:+1:

PR #719, opened 2022-04-14
Allow passing `PATRONI_LOG_*` environment variables to patroni process.

### LINKED ISSUE #718: PATRONI_LOG_* environment variables are ignored.
It seems that most environment variables are removed before starting the patroni process. I was attempting to configure patroni to log to a directory (instead of stdout) by using the `PATRONI_LOG_DIR` environment variable, but the `runit/patroni/run` script seems to be working against me here:

```
# Only small subset of environment variables is allowed. We don't want accidentally disclose sensitive information
for E in $(printenv -0 | tr '\n' ' ' | sed 's/\x00/\n/g' | grep -vE '^(KUBERNETES_(SERVICE|PORT|ROLE)[_=]|((POD_(IP|NAMESPACE))|HOSTNAME|PATH|PGHOME|LC_ALL|ENABLE_PG_MON)=)' | sed 's/=.*//g'); do
    unset $E
done
```
ref: https://github.com/zalando/spilo/blob/master/postgres-appliance/runit/patroni/run#L24-L26

Would it be possible to add `|(PATRONI_LOG_.*=)` to the regex to allow PATRONI_LOG_* variables to be passed to the process?

mhaley-tignis:
Quick change to the patroni run script to allow passing just patroni log config to  the patroni process. This configuration should not have any sensitive information, so should be fine to pass along.

fixes https://github.com/zalando/spilo/issues/718.

### DIFF
--- a/postgres-appliance/runit/patroni/run
+++ b/postgres-appliance/runit/patroni/run
@@ -21,7 +21,7 @@ fi
 ulimit -c unlimited
 
 # Only small subset of environment variables is allowed. We don't want accidentally disclose sensitive information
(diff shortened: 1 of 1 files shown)


mhaley-tignis:
@hughcapet are you able to assist here?

PR #721, opened 2022-04-22
Fix demo build

CyberDem0n:
It needs at least one deb file to copy and it seems that pg_permission repo was renamed to pg_permissions (plural)


FxKu:
:+1:

CyberDem0n:
:+1:

PR #723, opened 2022-05-02
Fall back in case wal-g restore_command hangs

haslersn:
The `restore_command` might not terminate (or at least run for a very long time). This is the case, for instance, if wal-g is enabled and it cannot reach the archive location. See: https://github.com/wal-g/wal-g/issues/1198

This commit adds a timeout for it. If the timeout triggers, then the `restore_command` terminates with exit code 124. PostgreSQL will next try to restore WAL from the `pg_wal` directory, then from the primary server (if configured) and, if that fails, then again from the archive location. See: https://www.postgresql.org/docs/current/warm-standby.html#STANDBY-SERVER-OPERATION

This way, a standby server can, in many (but not all) cases, start even if the archive location is down.

I chose a default `WAL_RESTORE_TIMEOUT` of 16 seconds assuming a transfer rate of at least 1 MB/s. (Note that WAL files are of size 16 MB or less when compressed.) If a user restores from the archive location at a slower rate, then the user must set the `WAL_RESTORE_TIMEOUT` environment variable to an integer larger than `16`.

### DIFF
--- a/ENVIRONMENT.rst
+++ b/ENVIRONMENT.rst
@@ -46,6 +46,7 @@ Environment Configuration Settings
 - **WALE_BACKUP_THRESHOLD_MEGABYTES**: maximum size of the WAL segments accumulated after the base backup to consider WAL-E restore instead of pg_basebackup.
 - **WALE_BACKUP_THRESHOLD_PERCENTAGE**: maximum ratio (in percents) of the accumulated WAL files to the base backup to consider WAL-E restore instead of pg_basebackup.
 - **WALE_ENV_DIR**: directory where to store WAL-E environment variables
+- **WAL_RESTORE_TIMEOUT**: timeout (in seconds) for restoring a single WAL file (at most 16 MB) from the backup location, 0 by default. A duration of 0 disables the timeout.
 - **WAL_S3_BUCKET**: (optional) name of the S3 bucket used for WAL-E base backups.
 - **AWS_ACCESS_KEY_ID**: (optional) aws access key
 - **AWS_SECRET_ACCESS_KEY**: (optional) aws secret key
--- a/postgres-appliance/major_upgrade/inplace_upgrade.py
+++ b/postgres-appliance/major_upgrade/inplace_upgrade.py
@@ -52,7 +52,7 @@ def update_configs(new_version):
 
     # update wal-e/wal-g envdir files
     restore_command = shlex.split(config['postgresql'].get('recovery_conf', {}).get('restore_command', ''))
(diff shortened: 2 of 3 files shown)


CyberDem0n:
Well, while it could work, I don't like your assumptions about 1MB/s and default timeout 16s.
The default behavior should remain as it is, i.e. no timeout at all.

haslersn:
Alright, I changed the default to 0 (= disabled) and adjusted the commit message.

CyberDem0n:
There are two more places where the `restore_command` is used (standby cluster and clone): https://github.com/zalando/spilo/blob/2daefafc666da0c44ce1f030e65cc2d10ab3f7ed/postgres-appliance/scripts/configure_spilo.py#L183-L231

haslersn:
> There are two more places where the `restore_command` is used

I know, but I thought the other two places are used by patroni to bootstrap or clone a cluster. Or are they also given to PostgreSQL as is? In the latter case, I'd add the timeout there, too.

haslersn:
I force-pushed a change that reduces the line length, because CI complained.

Can you explain why both `bootstrap.clone_with_wale.recovery_conf` and `bootstrap.dcs.standby_cluster.restore_command` are both configured? I think the first is dynamic configuration and the second is local configuration. In this case, does local configuration overwrite dynamic configuration? If yes, then why is `bootstrap.clone_with_wale.recovery_conf` required?

CyberDem0n:
Have you read https://patroni.readthedocs.io/en/latest/replica_bootstrap.html? It covers custom bootstrap (clone) and standby cluster configuration. These are different features and in both cases they need to have the restore_command configured independantly.

CyberDem0n:
Now major upgrade tests are failing because the restore_command is looking different: https://github.com/zalando/spilo/blob/2daefafc666da0c44ce1f030e65cc2d10ab3f7ed/postgres-appliance/major_upgrade/inplace_upgrade.py#L55

haslersn:
Alright, I changed the order of envdir and timeout.

CyberDem0n:
I would still suggest the following change:
```diff
-    if len(restore_command) > 4 and restore_command[0] == 'envdir':
+    if len(restore_command) > 6 and restore_command[0] == 'envdir':
```

haslersn:
Makes sense with the current code, so I applied your patch.

haslersn:
Is this ready for approval?

FxKu:
:+1:

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you @haslersn!

PR #729, opened 2022-05-20
Bump walg to v2.0.0

Jasper-Ben:
It is worth noting that v2.0 Postgres delta backups are not backward-compatible
with the pre-2.0 versions.


sdudoladov:
:+1:

CyberDem0n:
:+1:

Issue #736, opened 2022-06-06
Found a vulnerability on dpkg package


sasadangelo:
Hi Team,

Recently my team found the following vulnerability:
```
CVE-2022-1664

   Policy Status
   Active

   Summary
   A malicious source package could write files outside the unpack directory.

   Vendor Security Notice IDs   Official Notice   
   usn-5446-1                   http://www.ubuntu.com/usn/usn-5446-1   

   Affected Packages   Policy Status   How to Resolve                         Security Notice   
   dpkg                Active          Upgrade dpkg to >= 1.19.0.5ubuntu2.4   usn-5446-1
```

My spilo version is currently frozen to Nov 11 + additional fixes done by me and not integrated in official spilo code (specially to include PgBackrest). I know I should upgrade it to the latest version, however, I am sure the problem is still present on the latest version. In all our docker image we solve this problem simply adding:

```apt-get upgrade -y```

in our Dockerfile. This will be useful also for Spilo because it always start from:

```
ARG BASE_IMAGE=ubuntu:18.04
...
FROM $BASE_IMAGE as builder-false
```

that is an image frozen. However, with time new vulnerabilities will be found and it is good practice to upgrade the OS distribution packages to the latest version. I tried to do it in my own version but I got the following problem:
```
ln: failed to create symbolic link '/usr/lib/postgresql/11/lib/postgis-2.5.so': File exists
```

and I am not sure if the fix with ```ln -sf ...``` is appropriate.
Anyone can help?

The help I need is the following:
1. understand if it is possible to add ```apt-get upgrade -y``` in the official code.
2. if the problem I found it is safe to solve with ```ln -sf ...``` forcing the link even if the file already exists.

Thank you in advance for your help.

CyberDem0n:
> Found a vulnerability on dpkg package

What are the chances that it could be exploited externally?

> In all our docker image we solve this problem simply adding:

> apt-get upgrade -y

Yeah, currently it is missing in the Dockerfile, in the hope the latest ubuntu image is up-to-date.

> Spilo because it always start from:
>ARG BASE_IMAGE=ubuntu:18.04
...
FROM $BASE_IMAGE as builder-false
>that is an image frozen

It is not frozen, you can either build with `--pull`, so it will pull the latest image or do `docker pull` before the build. 

> I tried to do it in my own version but I got the following problem:

> ln: failed to create symbolic link '/usr/lib/postgresql/11/lib/postgis-2.5.so': File exists

[...]

Issue #726, opened 2022-05-12
Providing Patroni with a custom CA bundle


melnik13:
Hello,

I use [zalando/posrgres-operator](https://github.com/zalando/postgres-operator) that utilizes Spilo image.

Is there a way to provide Spilo with a custom CA certificate to be passed to Patroni?

Patroni handles the `PATRONI_KUBERNETES_CACERT` environment variable, but Spilo discards this variable and it's never been passed to Patroni.

Here's the link to the same issue discussed in the [zalando/posrgres-operator](https://github.com/zalando/postgres-operator) repository: https://github.com/zalando/postgres-operator/issues/1877.

Is there some way to make Spilo passing thie CA certificate of Kubernetes API Server to Patroni?

Thank you!

CyberDem0n:
Yeah... Spilo currently discards all Patroni environment variables because it is relying on its own.
The `get_dcs_config()` function should be a bit changed so it handles `KUBERNETES_*` env variables similarly to remaining DCS: https://github.com/zalando/spilo/blob/22c9247cb9c8d226fe16545673bbc6d419f4d07c/postgres-appliance/scripts/configure_spilo.py#L691-L726

PR #737, opened 2022-06-07
Patroni 2.1.4

CyberDem0n:
- ensure all packages are upgraded to the latest version (apt-get -y upgrade)
- allow starting etcd on non-x86 platforms
- refactor `get_dcs_config()` function

Close https://github.com/zalando/spilo/issues/736, https://github.com/zalando/spilo/issues/726

### DIFF
--- a/postgres-appliance/scripts/configure_spilo.py
+++ b/postgres-appliance/scripts/configure_spilo.py
@@ -34,7 +34,7 @@ PROVIDER_UNSUPPORTED = "unsupported"
 USE_KUBERNETES = os.environ.get('KUBERNETES_SERVICE_HOST') is not None
 KUBERNETES_DEFAULT_LABELS = '{"application": "spilo"}'
 MEMORY_LIMIT_IN_BYTES_PATH = '/sys/fs/cgroup/memory/memory.limit_in_bytes'
-PATRONI_DCS = ('zookeeper', 'exhibitor', 'consul', 'etcd3', 'etcd')
+PATRONI_DCS = ('kubernetes', 'zookeeper', 'exhibitor', 'consul', 'etcd3', 'etcd')
 AUTO_ENABLE_WALG_RESTORE = ('WAL_S3_BUCKET', 'WALE_S3_PREFIX', 'WALG_S3_PREFIX', 'WALG_AZ_PREFIX', 'WALG_SSH_PREFIX')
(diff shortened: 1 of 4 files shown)


postgres-appliance/scripts/configure_spilo.py
```diff
@@ -689,39 +697,38 @@ def pystache_render(*args, **kwargs):
 
 
 def get_dcs_config(config, placeholders):
+    # (KUBERNETES|ZOOKEEPER|EXHIBITOR|CONSUL|ETCD3|ETCD)_(HOSTS|HOST|PORT|...)
+    dcs_configs = defaultdict(dict)
+    for name, value in placeholders.items():
+        if '_' not in name:
+            continue
+        dcs, param = name.lower().split('_', 1)
+        if dcs in PATRONI_DCS:
```
sdudoladov, line 706: should not that raise a warning if the condition does not hold ?

postgres-appliance/scripts/configure_spilo.py
```diff
@@ -647,9 +646,18 @@ def get_placeholders(provider):
...
+    cgroup_v2_memory_limit_path = '/sys/fs/cgroup/memory.max'
+
+    if os.path.exists(cgroup_memory_limit_path):
+        with open(cgroup_memory_limit_path) as f:
             os_memory_mb = int(f.read()) / 1048576
+    elif os.path.exists(cgroup_v2_memory_limit_path):
+        with open(cgroup_v2_memory_limit_path) as f:
+            try:
+                os_memory_mb = int(f.read()) / 1048576
+            except Exception:  # there could be "max"
```
sdudoladov, line 659: what does the phrase 
>  there could be "max"

mean here ?

To me the code reads like `If there is any exception, pretend there is no limit on os_memory_mb by setting it to the max possible value`

postgres-appliance/scripts/configure_spilo.py
CyberDem0n: No, there could be many different environment variables, this is absolutely normal.

postgres-appliance/scripts/configure_spilo.py
sdudoladov: oh now I understand.
can you change the comment to 

>  string literal "max" is a possible value

?

FxKu:
:+1:

CyberDem0n:
:+1:

PR #742, opened 2022-06-24
Enable build and push of spilo-cdp-15

hughcapet:
Until final release the code will live in the feature/postgresql-15 branch


CyberDem0n:
:+1:

PR #748, opened 2022-07-01
Revert changes in upgrade scripts

hughcapet:
base upgrade output directory is now removed  after a successful pg_upgrade --check run


hughcapet:
👍

Issue #754, opened 2022-07-12
Allow setting primary_slot_name for standby_cluster


machine424:
Hello,

According to https://patroni.readthedocs.io/en/latest/SETTINGS.html:

> primary_slot_name: which slot on the remote master to use for replication. This parameter is optional, the default value is derived from the instance name (see function slot_name_from_member_name).

I don't think the default value is always set. If I follow `slot_name_from_member_name`:

https://github.com/zalando/patroni/blob/5b1fd2377647d81844d56b8a1f7fe574f39e155a/patroni/postgresql/config.py#L542-L545

On our standby cluster, `if use_slots and not (is_remote_master and member.no_replication_slot):` can never be true if we do not provide a `primary_slot_name`.

Am I missing something?

PR #755, opened 2022-07-12
Make it possible to set a slot for standby cluster (#754)

machine424:
Addresses https://github.com/zalando/spilo/issues/754

### DIFF
--- a/postgres-appliance/scripts/configure_spilo.py
+++ b/postgres-appliance/scripts/configure_spilo.py
@@ -188,6 +188,9 @@ bootstrap:
       {{#STANDBY_PORT}}
       port: {{STANDBY_PORT}}
       {{/STANDBY_PORT}}
+      {{#STANDBY_PRIMARY_SLOT_NAME}}
+      primary_slot_name: {{STANDBY_PRIMARY_SLOT_NAME}}
+      {{/STANDBY_PRIMARY_SLOT_NAME}}
     {{/STANDBY_CLUSTER}}
     ttl: 30
     loop_wait: &loop_wait 10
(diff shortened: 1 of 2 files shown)


postgres-appliance/scripts/configure_spilo.py
```diff
@@ -188,6 +188,9 @@ def deep_update(a, b):
       {{#STANDBY_PORT}}
       port: {{STANDBY_PORT}}
       {{/STANDBY_PORT}}
+      {{#STANDBY_SLOT}}
+      primary_slot_name: {{STANDBY_SLOT}}
+      {{/STANDBY_SLOT}}
```
CyberDem0n, line 193: I think this one is a bit more consistent with remaining env variable names:
```suggestion
      {{#STANDBY_PRIMARY_SLOT_NAME}}
      primary_slot_name: {{STANDBY_PRIMARY_SLOT_NAME}
      {{/STANDBY_PRIMARY_SLOT_NAME}}
```

CyberDem0n (commented):
Please also add the variable to the https://github.com/zalando/spilo/blob/39e41fb5f781498a50ab8edddec0ce2e0d75168c/postgres-appliance/scripts/configure_spilo.py#L611-L613 and update the documentation.

postgres-appliance/scripts/configure_spilo.py
CyberDem0n: No, I am talking about Spilo docs: https://github.com/zalando/spilo/blob/master/ENVIRONMENT.rst

machine424: I saw that the other `STANDBY_*` were not documented, I added some of them. (even though the best doc is configure_spilo.py ;) )

I'm stubborn, but do you really think patroni doc is ok?

https://patroni.readthedocs.io/en/latest/replica_bootstrap.html

> You can use Patroni’s permanent replication slots feature on the primary cluster to maintain a replication slot with the same name as primary_slot_name, or its default value if primary_slot_name is not provided.

https://patroni.readthedocs.io/en/latest/SETTINGS.html

> primary_slot_name: which slot on the remote master to use for replication. This parameter is optional, the default value is derived from the instance name (see function slot_name_from_member_name).

hughcapet:
👍

CyberDem0n:
:+1:

postgres-appliance/scripts/configure_spilo.py
machine424: Hi,

I didn't forget about this @CyberDem0n, I just need a yes or no: )

PR #760, opened 2022-07-19
Feature/ timeline id in clone section for Spilo

Sudeepta92:
To modify Spilo image to process timeline id in the clone section.

### DIFF
--- a/postgres-appliance/bootstrap/clone_with_wale.py
+++ b/postgres-appliance/bootstrap/clone_with_wale.py
@@ -27,9 +27,12 @@ def read_configuration():
                         dest='recovery_target_time_string')
     parser.add_argument('--dry-run', action='store_true', help='find a matching backup and build the wal-e '
                         'command to fetch that backup without running it')
+    parser.add_argument('--recovery-target-timeline',
+                        help='the timeline up to which recovery will proceed',
+                        dest='recovery_target_timeline_string')
     args = parser.parse_args()
 
-    options = namedtuple('Options', 'name datadir recovery_target_time dry_run')
+    options = namedtuple('Options', 'name datadir recovery_target_time recovery_target_timeline dry_run')
     if args.recovery_target_time_string:
         recovery_target_time = parse(args.recovery_target_time_string)
         if recovery_target_time.tzinfo is None:
@@ -37,7 +40,12 @@ def read_configuration():
     else:
         recovery_target_time = None
 
-    return options(args.scope, args.datadir, recovery_target_time, args.dry_run)
+    if args.recovery_target_timeline_string == None:
+        recovery_target_timeline = "latest"
+    else:
+        recovery_target_timeline = args.recovery_target_timeline_string
+
+    return options(args.scope, args.datadir, recovery_target_time, recovery_target_timeline, args.dry_run)
 
 
 def build_wale_command(command, datadir=None, backup=None):
@@ -65,7 +73,7 @@ def fix_output(output):
             yield '\t'.join(line.split())
 
 
-def choose_backup(backup_list, recovery_target_time):
+def choose_backup(backup_list, recovery_target_time, recovery_target_timeline):
     """ pick up the latest backup file starting before time recovery_target_time"""
 
     match_timestamp = match = None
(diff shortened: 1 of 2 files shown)


postgres-appliance/bootstrap/clone_with_wale.py
```diff
@@ -65,7 +67,7 @@ def fix_output(output):
             yield '\t'.join(line.split())
 
 
-def choose_backup(backup_list, recovery_target_time):
+def choose_backup(backup_list, recovery_target_time, recovery_target_timeline):
```
hughcapet, line 70: How do you use this newly added parameter to choose the latest backup file? It seems `recovery_target_timeline` argument is not used by the function

Sudeepta92: Hey, yes - so we wanted to add a new parameter "timeline id" for the Zalando Operator and Spilo. And then we realised there is a bug that Spilo always restores to the backup closest to the provided timestamp irrespective of the timeline. Which means that its possible that a restore is done from an older timeline if the backup closest to the timestamp was on that timeline - which would be a wrong scenario in case restore is required from the latest timeline. So apart from the current PRs I will create a bug on this and open a PR to it.

hughcapet (changes requested):

CyberDem0n:
@Sudeepta92 let's step back and explain what problem are you trying to solve?

The thing is that **recovery_target_timeline** has only very narrow use: https://www.postgresql.org/docs/current/continuous-archiving.html#BACKUP-TIMELINES

>  The default behavior of recovery is to recover to the latest timeline found in the archive. **If you wish to recover to the timeline that was current when the base backup was taken or into a specific child timeline (that is, you want to return to some state that was itself generated after a recovery attempt), you need to specify current or the target timeline ID in [recovery_target_timeline](https://www.postgresql.org/docs/current/runtime-config-wal.html#GUC-RECOVERY-TARGET-TIMELINE)**. You cannot recover into timelines that branched off earlier than the base backup.

This is exactly what postgres-operator avoids: every time you deploy a new cluster it gets a new and unique archive location. That is, we will never have the situation when two recovery attempts are performed with the same archive location, what makes this feature effectively useless...

thedatabaseme:
> This is exactly what postgres-operator avoids: every time you deploy a new cluster it gets a new and unique archive location. That is, we will never have the situation when two recovery attempts are performed with the same archive location, what makes this feature effectively useless...

@CyberDem0n I'm not sure if I understand you correctly. It's not always true that a unique archive location is created for a database. This can be because of two reasons. First, you can specify that there should be no prefix and suffix within the archive location path (S3), which leads to a location which only holds the cluster name and version and **no** UID. Second, you can do an "in place restore" (described [here](https://github.com/zalando/postgres-operator/blob/89375186b3fd3df2e787f9db5a85c13166729581/docs/user.md#restore-in-place)) This will lead to a new cluster UID but in the case above where the UID is not represented in the archive location, this will lead to the same problem.

On the Operator repository, there is also a open PR to add the needed configuration to the postgresql manifest, they seem to be willing to implement this. (See [here](https://github.com/zalando/postgres-operator/pull/1961))

When having UIDs in the backup location path, this makes it nearly impossible to automate a restore process and make backups selectable over some kind of webinterface, cause you probably will not have a history of all UIDs that a database might had in the past, therefore you cannot make the backups selectable anymore using an automation. I think this is the reason why one would disable the prefix / suffix stuff.

Just my thoughts.

Philip

CyberDem0n:
> you can specify that there should be no prefix and suffix within the archive location path (S3), which leads to a location which only holds the cluster name and version and no UID.

Well, if you shoot your own foot, it hurts badly. Just don't do it. Two different clusters must not be writing to the same archive location.

thedatabaseme:
> > you can specify that there should be no prefix and suffix within the archive location path (S3), which leads to a location which only holds the cluster name and version and no UID.
> 
> Well, if you shoot your own foot, it hurts badly. Just don't do it. Two different clusters must not be writing to the same archive location.

They are not different clusters, just a restored one which takes place from the original one (inplace restore). And second, I don't understand why you use an offensive term like "shoot in the foot", it's the concept of timelines in Postgres, so not a shady workaround. 

Philip

CyberDem0n:
If the old cluster is still up and running and the new cluster is forked from the old one (with the new timeline), they are two different clusters.
They just have two common things:
1. cluster system identifier
2. and the ancestor

If you are unlucky enough, both clusters will end up on the same timeline and comparable LSNs, and there will be clashes in WAL file names.

Even if the old cluster is down, there will be a mess in the archive after PITR.

I don't understand the offensiveness of the "shoot your own foot". It is a well-known idiom and in this context, it means that we shouldn't be creating problems in the first place in order to make a significant effort to solve these problems later.

PR #762, opened 2022-07-20
Disable wal_archiving during major upgrade

CyberDem0n:
- We don't want WAL files to be archived to the location used by the old cluster (old version).
- bump bg_mon commit id.


hughcapet:
👍

CyberDem0n:
:+1:

PR #771, opened 2022-09-03
Prefix array_eq with schema to avoid calling overwritten operator.

mneverov:
Previously, a regular user could get superuser role by overwriting `array_eq` operator in public schema:
```sql
create function public.array_eq(text[], text[]) returns boolean
language sql
as $_$
alter user <regular_user> with superuser;
select pg_catalog.array_eq($1, $2);
$_$;
```

When a major version upgrade is triggered or PiTR, the `post_init.sh` is called and since `=` operator is not prefixed with `pg_catalog` the more specific overwritten operator from public schema is called, granting `superuser` role.
More details in [pganalyze blog](https://pganalyze.com/blog/5mins-postgres-security-patch-releases-pgspot-pghostile).
I used [pghostile](https://github.com/aiven/pghostile) to overwrite multiple operators/functions to test.

### DIFF
--- a/postgres-appliance/scripts/post_init.sh
+++ b/postgres-appliance/scripts/post_init.sh
@@ -51,7 +51,7 @@ GRANT SELECT ON TABLE public.pg_auth_mon TO robot_zmon;
 CREATE EXTENSION IF NOT EXISTS pg_cron SCHEMA public;
 DO \$\$
 BEGIN
-    PERFORM 1 FROM pg_catalog.pg_proc WHERE pronamespace = 'cron'::pg_catalog.regnamespace AND proname = 'schedule' AND proargnames = '{p_schedule,p_database,p_command}';
+    PERFORM 1 FROM pg_catalog.pg_proc WHERE pronamespace = 'cron'::pg_catalog.regnamespace AND proname = 'schedule' AND proargnames OPERATOR(pg_catalog.=) '{p_schedule,p_database,p_command}';
     IF FOUND THEN
         ALTER FUNCTION cron.schedule(text, text, text) RENAME TO schedule_in_database;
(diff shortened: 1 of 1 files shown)


hughcapet:
Thank you for raising this! By the way, why do you choose to specify an operator's schema explicitly but not setting search_path at the beginning? It seems we should probably extend this change to other operators in the script (or am I missing something?).

hughcapet:
> I tried to do that (set search_path in the beginning of the script and reset it in the end), but the script failed.

Could you please share the error that was raised then?

mneverov:
> Could you please share the error that was raised then?

No, I don't have the logs unfortunately. I set the path & ran the tests  - they succeeded, so looks like it works.

postgres-appliance/scripts/post_init.sh
```diff
@@ -144,7 +146,8 @@ CREATE TABLE IF NOT EXISTS public.postgres_log (
     application_name text,
     CONSTRAINT postgres_log_check CHECK (false) NO INHERIT
 );
-GRANT SELECT ON public.postgres_log TO admin;"
+GRANT SELECT ON public.postgres_log TO admin;
+RESET search_path;"
```
CyberDem0n, line 150: Why do you want to reset it?

IMO, we better use `PGOPTIONS` environment variable to globally enforce `search_path` for all psql connections.
There is already one place where `PGOPTIONS` is used: https://github.com/zalando/spilo/blob/dd7f7181b733ed69317a9ecaa780b5563bac8ee5/postgres-appliance/scripts/post_init.sh#L206

From there it should be moved to the beginning of the script and ` -c search_path=pg_catalog` should be added to the value.

PR #773, opened 2022-09-15
Drop pg_repack before upgrade --check

hughcapet:
There are recent [commits](https://github.com/postgres/postgres/commit/09878cdd489ff7aca761998e7cb104f4fd98ae02) into PG14+ that add an additional check for user-defined objects that refer to internal polymorphic functions with arguments of type "anyarray" or "anyelement". It fails the pg_upgrade check in case such objects are found and requires its deletion first.
pg_repack used to have such functions prior to v1.4.7


hughcapet:
👍

PR #777, opened 2022-09-22
Multi arch image ci

DavidSpek:
It appears that the Zalando registry doesn’t support multi-arch images, which is a big problem for people using mixed architecture nodes in a cluster. This PR adds a workflow that will build a multi-arch image when a new tag is pushed to the repo and push it to GHCR, solving the problem for users with mixed architecture nodes.

The image created by this workflow can be tested at https://github.com/pluralsh/spilo/pkgs/container/spilo-14.

@hughcapet If this workflow is alright with you, I’d be great to have it merged before the new release tag is pushed today.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -20,7 +20,7 @@ jobs:
       with:
         # list of Docker images to use as base name for tags
         images: |
-          ghcr.io/pluralsh/spilo-14
+          ghcr.io/zalando/spilo-14
         # generate Docker tags based on the following events/attributes
         tags: |
           type=ref,event=tag
@@ -43,17 +43,3 @@ jobs:
         tags: ${{ steps.meta.outputs.tags }}
         labels: ${{ steps.meta.outputs.labels }}
         platforms: linux/amd64,linux/arm64
-  # release:
-  #   name: Create GitHub release
-  #   runs-on: ubuntu-latest
-  #   needs: publish
-  #   permissions:
-  #     contents: write
-  #     discussions: write
-  #   steps:
-  #   - name: Checkout
-  #     uses: actions/checkout@v3
-  #   - name: Release
-  #     uses: softprops/action-gh-release@v1
-  #     with:
-  #       generate_release_notes: true
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,59 @@
+name: Publish container
+on:
+  push:
+    tags:
+      - '*'
+jobs:
+  publish:
+    name: Build and push Spilo container
+    runs-on: ubuntu-latest
+    permissions:
+      contents: 'read'
+      id-token: 'write'
+      packages: 'write'
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+    - name: Docker meta
+      id: meta
+      uses: docker/metadata-action@v4
+      with:
+        # list of Docker images to use as base name for tags
+        images: |
+          ghcr.io/pluralsh/spilo-14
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -0,0 +1,45 @@
...
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+    - name: Docker meta
+      id: meta
+      uses: docker/metadata-action@v4
+      with:
+        # list of Docker images to use as base name for tags
+        images: |
+          ghcr.io/zalando/spilo-14
```
CyberDem0n, line 23: Could it somehow be dynamic? I.e. extract the `PGVERSION` value from the Dockerfile and have something like `ghcr.io/zalando/spilo-$PGVERSION`.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,45 @@
...
+      with:
+        registry: ghcr.io
+        username: ${{ github.repository_owner }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+    - name: Build and push
+      uses: docker/build-push-action@v3
+      with:
+        context: "postgres-appliance/"
+        file: "postgres-appliance/Dockerfile"
+        push: true
```
CyberDem0n, line 42: Before pushing we would like to test at least amd64 image, like it is done here: https://github.com/zalando/spilo/blob/71491609ae88180655afdbb76fb8692f612473b7/.github/workflows/tests.yaml#L29-L30

DavidSpek: I'll try adding that. Would need to test the action with a delayed push.

.github/workflows/publish.yml
CyberDem0n:
```sh
sed -n 's/^ARG PGVERSION=\([1-9][0-9]*\).*$/\1/p' Dockerfile
```

.github/workflows/publish.yml
hughcapet: Any news here, @DavidSpek?

mmoscher:
@CyberDem0n and @hughcapet are you really interested and willing in hosting additional, multiarch, images on ghcr?

Your build pipeline already includes building and publishing arm64 images into your own registry (https://github.com/pluralsh/spilo/blob/d518588c15c23f9711830f43d47f88c8b2130e72/delivery.yaml#L63).
Thus, one can already use these images, eg. for running the postgres-operator on an aarch64/arm64 machine.

If you're happy to see this PR rise, I would take over the idea and file a new PR.
What's your opinion?

mmoscher:
//Follow-up: this kind of workflow could also be interesting for the postgres-operator itself. It would - for now - solve your internal limitation, eg. hosting multiarch container, and would enable others (the community) to more easily run spilo/postgres-operator under arm64/aarch64. Cf. https://github.com/zalando/postgres-operator/issues/1940, https://github.com/zalando/postgres-operator/issues/1109, https://github.com/zalando/postgres-operator/issues/2030

hughcapet:
> are you really interested and willing in hosting additional, multiarch, images on ghcr?

yes, we are. It will be nice it you will bring this idea to the completion

mmoscher:
@hughcapet here we go: https://github.com/zalando/spilo/pull/790
Waiting for the arm64 build to complete (https://github.com/mmoscher/spilo/actions/runs/3245461373/jobs/5323031631). Afterwards it should be ready for review.

DavidSpek:
@mmoscher thanks for picking this up, I got stuck with other obligations but happy that this has been merged now.

PR #787, opened 2022-10-07
Add more extensions for PG15

hughcapet:
- add pg-checksums and pgq-node extensions for pg15
- don't patch pgextwlist


FxKu:
:+1:

hughcapet:
👍

PR #769, opened 2022-08-30
Pass AWS EKS IRSA environment variables to WAL-G

avbasov:
When one uses the IAM Roles for Service Accounts feature for assigning an IAM role with S3 access to a pod, EKS mounts an auth token and sets environment variables in the pod. Everything works fine except the archive command.
So I've added those vars to the WAL-G environment.

PR #789, opened 2022-10-13
Pass AWS STS regional endpoints environment variable to WAL-G

ThomasK33:
While #769 started passing the EKS IRSA environment variable to WAL-G, it omitted to add the `AWS_STS_REGIONAL_ENDPOINTS` environment variable, which is also configured automatically in EKS clusters using IRSA.

This variable becomes especially important and required in scenarios where EKS clusters are run in a VPC without an (Egress-Only) Internet Gateway and cannot reach the global STS endpoint.

AWS docs reference: https://docs.aws.amazon.com/eks/latest/userguide/configure-sts-endpoint.html

This would fix zalando/postgres-operator#1124 for clusters depending on a regional STS endpoint.


Jan-M:
:+1:

hughcapet:
👍

hughcapet:
Merged. Thank you, @ThomasK33!

PR #664, opened 2021-11-22
Add build support on non-amd64 platforms

CyberDem0n:
the biggest problem was the lack of non-amd64 binaries of wal-g which we solve by introducing yet another stage.

So far it was tested only on arm64.

PR #790, opened 2022-10-13
adds workflow for multiarch builds (linux/amd64, linux/arm64) being published to ghcr.io

mmoscher:
### Changelog

- introduces new GHA workflow to build and publish spilo multiarch images on ghcr.io. Workflow is triggered on tag creation.
- currently linux/amd64 and linux/arm64 images are built
- linux/amd64 build is tested using postgres-appliance/tests/test_spilo.sh
- **Note** the linux/arm64 takes much more time, due to emulation, then the linux/amd64 build (7 vs. 120min)

### Changes

- adds GHA workflow .github/workflows/publish-ghcr-container.yaml
- extends postgres-appliance/tests/docker-compose.yml to allow spilo image definition using environment variable, cf. postgres-appliance/tests/README
- adds small section in the README mentioning that multiarch builds can be found on ghcr.io

### Tests

Tests were done by tagging the forked repo. Tag `p1-2` was successful.

- workflow run can be found here: https://github.com/mmoscher/spilo/actions/runs/3245461373/jobs/5323031631
- multiarch images can be found here:

### Follow-Up:

- wal-g multiarch builds are now available. Thus, https://github.com/zalando/spilo/pull/664 can be reverted and additional self-builds can be omitted. Finally, the arm64 build-time should decrease enormously.

### DIFF
--- a/.github/workflows/publish-ghcr-container.yaml
+++ b/.github/workflows/publish-ghcr-container.yaml
@@ -1,0 +1,65 @@
+name: Publish multiarch images on ghcr.io
+on:
+  push:
+    tags:
+      - '*'
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
+
+jobs:
+  publish:
+    name: Build and push Spilo multiarch images
+    runs-on: ubuntu-latest
+    permissions:
+      contents: 'read'
+      packages: 'write'
+    defaults:
+      run:
+        shell: bash
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+
+    - name: Derive spilo image name
+      id: image
+      working-directory: postgres-appliance
+      run: |
+        PGVERSION=$(sed -n 's/^ARG PGVERSION=\([1-9][0-9]*\).*$/\1/p' Dockerfile)
+        IMAGE="${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}-$PGVERSION:${GITHUB_REF/refs\/tags\//}"
+        echo "::set-output name=NAME::$IMAGE"
+
+    - name: Set up QEMU
+      uses: docker/setup-qemu-action@v2
+
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v2
+
+    - name: Login to GHCR
+      uses: docker/login-action@v2
+      with:
+        registry: ${{ env.REGISTRY }}
+        username: ${{ github.actor }}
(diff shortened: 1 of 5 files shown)


mmoscher:
@CyberDem0n can you have a look on the follow-up? If you think it's a valid solution, I'll file another PR.

.github/workflows/publish-ghcr-container.yaml
```diff
@@ -0,0 +1,73 @@
...
+    - name: Install flake8 and docker-compose
+      run: python -m pip install flake8 docker-compose==1.17.1
+
+    - name: Derive spilo image name
+      id: image
+      working-directory: postgres-appliance
+      run: |
+        PGVERSION=$(sed -n 's/^ARG PGVERSION=\([1-9][0-9]*\).*$/\1/p' Dockerfile)
+        IMAGE="${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}-$PGVERSION:${GITHUB_REF/refs\/tags\//}"
+        echo "::set-output name=NAME::$IMAGE"
```
hughcapet, line 39: Can we switch to [the new format](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter) here? The `set-output` command is deprecated.

hughcapet:
As for the follow-up, it is a valid solution but only unless [feature/postgresql-15](https://github.com/zalando/spilo/tree/feature/postgresql-15) is merged. We have recently migrated to Ubuntu 22.04 there and will actually build wal-g for both platforms.

.github/workflows/publish-ghcr-container.yaml
mmoscher: Nice :), didn't know about it yet. Looks interesting!

mmoscher:
> As for the follow-up, it is a valid solution but only unless [feature/postgresql-15](https://github.com/zalando/spilo/tree/feature/postgresql-15) is merged. We have recently migrated to Ubuntu 22.04 there and will actually build wal-g for both platforms.

thanks for the info. To this end it sounds more like unnecessary work and I would leave it as is

mmoscher:
@hughcapet suggested changes are made. Please have a re-review :)

Jan-M:
:+1:

@mmoscher von Aachener zu Aachener - great work. thanks for helping out.

hughcapet:
👍

mmoscher:
Thank's @Jan-M & @hughcapet  =)

PR #786, opened 2022-10-07
Do not patch pgextwlist extension

hughcapet:
Pgextwlist doesn't require patching anymore (see https://github.com/dimitri/pgextwlist/commit/b440748a481ebba8f017bd54063af9bb11e25708)

PR #800, opened 2022-11-14
Install pgextwlist unconditionally

hughcapet:
Fixes the regression introduced in https://github.com/zalando/spilo/pull/786 - pgextwlist should be also installed in DEMO builds.


Jan-M:
:+1:

PR #802, opened 2022-11-16
PyYAML 6.0 compatibility

ants:
Version 6.0 of PyYAML requires a Loader parameter for yaml.load.

Defaulting to SafeLoader for good hygiene, previous behavior was to use FullLoader.


hughcapet:
Maybe we better switch to `safe_load()` then?

Issue #805, opened 2022-11-23
Remove curl at the end of Docker build


mbanck:
Spilo leaves curl (needed to download external projects/extensions during `docker build` installed after the build.

This is an unneeded attack surface in my opinion and there seems to be no problem in removing it:

```
docker exec -it  4acfce812fb4 bash
 ____        _ _
/ ___| _ __ (_) | ___
\___ \| '_ \| | |/ _ \
 ___) | |_) | | | (_) |
|____/| .__/|_|_|\___/
      |_|
[...]
root@4acfce812fb4:/home/postgres# apt remove curl
Reading package lists... Done
Building dependency tree       
Reading state information... Done
The following packages will be REMOVED:
  curl
0 upgraded, 0 newly installed, 1 to remove and 0 not upgraded.
After this operation, 397 kB disk space will be freed.
Do you want to continue? [Y/n] y
(Reading database ... 29333 files and directories currently installed.)
Removing curl (7.58.0-2ubuntu3.21) ...
root@4acfce812fb4:/home/postgres# 
```

hughcapet:
> needed to download external projects/extensions during docker build


well, not only
```
spilo % grep -r "curl" postgres-appliance/scripts 
postgres-appliance/scripts/wal-e-wal-fetch.sh:    INSTANCE_PROFILE=$(curl -s "$CREDENTIALS_URL")
postgres-appliance/scripts/wal-e-wal-fetch.sh:    source <(curl -s "$CREDENTIALS_URL$INSTANCE_PROFILE" | jq -r '"AWS_SECURITY_TOKEN=\"" + .Token + "\"\nAWS_SECRET_ACCESS_KEY=\"" + .SecretAccessKey + "\"\nAWS_ACCESS_KEY_ID=\"" + .AccessKeyId + "\""')
postgres-appliance/scripts/wal-e-wal-fetch.sh:    AZ=$(curl -s http://169.254.169.254/latest/meta-data/placement/availability-zone)
postgres-appliance/scripts/wal-e-wal-fetch.sh:    if curl -s "https://$HOST$FILE" "${TOKEN_HEADER[@]}" -H "x-amz-content-sha256: $EMPTYHASH" -H "x-amz-date: $TIME" \
postgres-appliance/scripts/patroni_wait.sh:    while [ "$(curl -so /dev/null -w '%{http_code}' "http://localhost:8008/$ROLE")" != "200" ]; do
[...]

PR #806, opened 2022-11-23
Remove curl at the end of Docker build (#805)

mbanck:
Remove curl at the end of Docker build


mbanck:
Retracting this PR as the underlying issue is no longer relevant either

PR #818, opened 2022-12-15
Add extension timescaledb-toolkit

ShutterQuick:
[TimescaleDB Toolkit](https://github.com/timescale/timescaledb-toolkit) adds a bunch of useful features (like e.g. downsampling) to TimescaleDB.
It is licensed under Timescale License (TSL), so I've set it up to only compile only when `TIMESCALEDB_APACHE_ONLY` isn't true.
Considering there's already TimescaleDB support in the image, I hope you'll also accept this as a natural companion extension.

Let me know if you'd like me to change anything in the MR and I'll update it.


ShutterQuick:
@hughcapet Sorry to at you directly, but could you let me know if this is something you'd be open for including at all.
If it's not, the PR might as well be closed. But if it is, I should keep rebasing it on top of master and keep it current.
And again, if it's something you're open for in principle, but not in the current form, let me know what to change.

hughcapet:
@ShutterQuick sorry for the delay, I will look into it shortly

ShutterQuick:
@hughcapet Thank you for the feedback!
I agree 100%. The reason I made it compile from source is that's what the script does for timescaledb, and I didn't stop to think that of course the reason for this is to support both the apache license and timescale license scenarios.
And yeah, agreed. The script for it is downright nasty with all the hoops necessary to jump through.

I've update the MR to instead install it from the timescale repo, which unsurprisingly makes it a lot saner.
Let me know if you want me to make any further changes.

postgres-appliance/build_scripts/base.sh
```diff
@@ -129,22 +129,41 @@ for version in $DEB_PG_SUPPORTED_VERSIONS; do
...
+                -DTAP_CHECKS=OFF -DPG_CONFIG="/usr/lib/postgresql/$version/bin/pg_config" \
+                -DAPACHE_ONLY="$TIMESCALEDB_APACHE_ONLY" -DSEND_TELEMETRY_DEFAULT=NO; then
+            make -C build install
+            strip /usr/lib/postgresql/"$version"/lib/timescaledb*.so
+            __timescaleDbInstalled="true"
+        fi
+        git reset --hard
+        git clean -f -d
+    done
+    popd
```
ShutterQuick, line 150: I wanted a check to make sure toolkit isn't installed if timescaledb isn't.
Got rid of the subshell in favour of using pushd/popd to be able to set the variable.

postgres-appliance/build_scripts/base.sh
```diff
@@ -129,22 +129,41 @@ for version in $DEB_PG_SUPPORTED_VERSIONS; do
...
+
+    if [ "${TIMESCALEDB_APACHE_ONLY}" != "true" ] && [ "${TIMESCALEDB_TOOLKIT}" = "true" ] && [ "${__timescaleDbInstalled}" = "true" ]; then
+        __versionCodename=$(sed </etc/os-release -ne 's/^VERSION_CODENAME=//p')
+        echo "deb [signed-by=/usr/share/keyrings/timescale_E7391C94080429FF.gpg] https://packagecloud.io/timescale/timescaledb/ubuntu/ ${__versionCodename} main" | tee /etc/apt/sources.list.d/timescaledb.list
+        curl -L https://packagecloud.io/timescale/timescaledb/gpgkey | apt-key --keyring /usr/share/keyrings/timescale_E7391C94080429FF.gpg add -
+
+        apt-get update
+        if [ $(apt-cache search --names-only "^timescaledb-toolkit-postgresql-${version}$" | wc -l) -eq 1 ]; then
+            apt-get install "timescaledb-toolkit-postgresql-$version"
```
ShutterQuick, line 159: Using `apt-cache` to check for the exact package name as I didn't want to swallow errors from `apt-get install`, by assuming errors from it means package not found.

I'm making the assumption here that if `apt-cache` outputs a single line that means the package is found.
I think this is okay, but it is a little nasty and if you prefer I can change from `wc -l` to instead grep for the package name in the output.

postgres-appliance/scripts/post_init.sh
```diff
@@ -179,6 +179,10 @@ while IFS= read -r db_name; do
     if [ "$UPGRADE_TIMESCALEDB" = "t" ]; then
         echo "ALTER EXTENSION timescaledb UPDATE;"
     fi
+    UPGRADE_TIMESCALEDB_TOOLKIT=$(echo -e "SELECT NULL;\nSELECT default_version != installed_version FROM pg_catalog.pg_available_extensions WHERE name = 'timescaledb_toolkit'" | psql -tAX -d "${db_name}" 2> /dev/null | tail -n 1)
+    if [ "$UPGRADE_TIMESCALEDB_TOOLKIT" = "t" ]; then
+        echo "ALTER EXTENSION timescaledb_toolkit UPDATE;"
+    fi
```
ShutterQuick, line 185: This is definitely monkey see monkey do. Copy-pasted from the lines above in the script so probably take an extra look at this.

ShutterQuick (commented):
Adding a few comments trying to anticipate review questions.

postgres-appliance/build_scripts/base.sh
hughcapet: regarding this `__timescaleDbInstalled` var...  I had a feeling timescale-toolkit can be used without timescaledb installed (with pure pg). Am I wrong?

ShutterQuick: I'm not sure! Haven't seen in said explicitly either way, guess I just didn't think it'd make much sense.
Want me to remove the check __timescaleDbInstalled?

postgres-appliance/build_scripts/base.sh
```diff
@@ -146,6 +146,22 @@ for version in $DEB_PG_SUPPORTED_VERSIONS; do
         done
     )
 
+    if [ "${TIMESCALEDB_APACHE_ONLY}" != "true" ] && [ "${TIMESCALEDB_TOOLKIT}" = "true" ]; then
+        __versionCodename=$(sed </etc/os-release -ne 's/^VERSION_CODENAME=//p')
```
hughcapet, line 150: maybe just `$(lsb_release -c -s)` ?

postgres-appliance/build_scripts/base.sh
```diff
@@ -146,6 +146,22 @@ for version in $DEB_PG_SUPPORTED_VERSIONS; do
...
+        echo "deb [signed-by=/usr/share/keyrings/timescale_E7391C94080429FF.gpg] https://packagecloud.io/timescale/timescaledb/ubuntu/ ${__versionCodename} main" | tee /etc/apt/sources.list.d/timescaledb.list
+        curl -L https://packagecloud.io/timescale/timescaledb/gpgkey | apt-key --keyring /usr/share/keyrings/timescale_E7391C94080429FF.gpg add -
```
hughcapet, line 152:
```
Warning: apt-key is deprecated. Manage keyring files in trusted.gpg.d instead (see apt-key(8)).
```
can you please switch to `gpg --dearmor`

postgres-appliance/build_scripts/base.sh
```diff
@@ -146,6 +146,22 @@ for version in $DEB_PG_SUPPORTED_VERSIONS; do
...
+
+        apt-get update
+        if [ $(apt-cache search --names-only "^timescaledb-toolkit-postgresql-${version}$" | wc -l) -eq 1 ]; then
+            apt-get install "timescaledb-toolkit-postgresql-$version"
```
ShutterQuick, line 156: Btw; this is a problem re. reproducible builds, in that the newest version for a given pgversion always will be selected. Is this a show-stopper?

I'm not quite what to do about this if it is, as it seems the (not so good, frankly) alternative is having an array of what exact toolkit version to install for what pgversion. What's your thoughts on this?

hughcapet: unfortunately, Spilo does not guarantee reproducible builds. it is reprodicible only when is pulled from a registry:)

postgres-appliance/build_scripts/base.sh
hughcapet: I think it is fine for now

postgres-appliance/build_scripts/base.sh
ShutterQuick: Yeah, true enough. What I mean is more that this dependency is probably likely to receive quite a bit more updates than most others and I'd imagine a high likelyhood of version changes for plugins on different builds of the same spilo release isn't great.

If you don't think it's a problem, I'm fine with keeping it as is, but if you think it's a problem I can add version fixing.

postgres-appliance/build_scripts/base.sh
ShutterQuick: I didn't use that because lsb_release isn't included in standard Ubuntu docker images.
Doesn't look like anything in the spilo image does either, so probably better just getting it from the text file.
Agree?

```
++ lsb_release -c -s
base.sh: line 150: lsb_release: command not found
```

postgres-appliance/build_scripts/base.sh
ShutterQuick: Sure thing! Done :)

postgres-appliance/build_scripts/base.sh
hughcapet: sorry, you're right. It is installed later during Patroni installation

FxKu:
:+1:

hughcapet:
👍

hughcapet:
Thank you for the contribution @ShutterQuick!

PR #821, opened 2022-12-27
Add USE_OLD_LOCALES test

hughcapet:
The newly added test verifies that with the old locales (Ubuntu 18.04) kept postgres sorting order is actually preserved on Ubuntu 22.04.
The test is only run on push to master branch.


FxKu:
:+1:

hughcapet:
👍

PR #824, opened 2023-01-03
add function to list last status of active cron jobs

FxKu:
We want to alert on failing cron jobs. But `robot_zmon` user has no rights to query tables cron.job and cron.job_run_details. Even when granted USAGE and SELECT, the details are protected by a [row level policy](https://github.com/zalando/spilo/blob/master/postgres-appliance/scripts/post_init.sh#L61). Therefore, the PR suggests to add a wrapper function with SECURITY DEFINER to allow the invoker to list the last status of active cron jobs.

### DIFF
--- a/postgres-appliance/scripts/metric_helpers.sql
+++ b/postgres-appliance/scripts/metric_helpers.sql
@@ -218,4 +218,29 @@ REVOKE ALL ON TABLE pg_stat_statements FROM public;
 GRANT SELECT ON TABLE pg_stat_statements TO admin;
 GRANT SELECT ON TABLE pg_stat_statements TO robot_zmon;
 
+CREATE FUNCTION get_last_status_active_cronjobs(
+  OUT jobid bigint,
+  OUT database text,
+  OUT command text,
+  OUT status text,
+  OUT return_message text,
+  OUT start_time timestamp with time zone,
+  OUT end_time timestamp with time zone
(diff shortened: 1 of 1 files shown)


hughcapet:
do we want this function to be accessible for all users or only for `robot_zmon`?

FxKu:
all users that have access to `metric_helpers` schema, I would say

hughcapet:
👍

postgres-appliance/scripts/metric_helpers.sql
```diff
@@ -193,29 +185,48 @@ FROM (
...
+       job_run_details.start_time,
+       job_run_details.end_time
+  FROM job
+  JOIN job_run_details USING (jobid)
+ WHERE job.active
+ ORDER BY job_run_details.jobid, job_run_details.start_time DESC NULLS LAST;
+$_$
+LANGUAGE sql SECURITY DEFINER STRICT SET search_path to 'cron';
+
+CREATE OR REPLACE VIEW status_active_cronjobs AS SELECT * FROM get_last_status_active_cronjobs();
```
hughcapet, line 222: can we please add 'last' to the name? as it is a bit confusing now - seems it should return not the last but all statuses

hughcapet:
👍

FxKu:
:+1:

PR #825, opened 2023-01-09
PyYAML 6.0 compatibility

hughcapet:
Switch to safe_load(), as load() is unsafe and since pyyaml 6.0 always requires loader to be specified


Jan-M:
:+1:

hughcapet:
👍

PR #826, opened 2023-01-09
Enforce search_path=pg_catalog

hughcapet:
- Enforce search_path=pg_catalog for post_init and wale_restore scripts
- call `pg_is_in_recovery()` with schema in several places

the problem was raised in https://github.com/zalando/spilo/pull/771


FxKu:
:+1:

hughcapet:
👍

PR #828, opened 2023-01-20
Don't try building timescaledb for PG15

hughcapet:
Even though timescaledb can now be compiled for PG15, it is not officially supported. We will disable its build for PG15 for now.


sdudoladov:
:+1:

PR #830, opened 2023-01-28
CDP multi-arch image build

hughcapet:
Build multi-arch image and push it to Zalando Container Registry on every commit to master


hughcapet:
@mmoscher it is for internal use only. OS will still be published in ghcr :)

mmoscher:
> @mmoscher it is for internal use only. OS version will still be published in ghcr :)

@hughcapet ok, good to know 😃! but also cool to see that you now have a multi-arch registry internally

FxKu:
:+1:

hughcapet:
👍

PR #842, opened 2023-02-10
Use relevant params in pgqd ini file

hughcapet:
Seems, config file didn't set anything for a long time (if ever did).

### DIFF
--- a/postgres-appliance/pgq_ticker.ini
+++ b/postgres-appliance/pgq_ticker.ini
@@ -1,9 +1,7 @@
 [pgqd]
-job_name = ticker
 initial_database = postgres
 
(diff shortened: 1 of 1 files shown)


mblsf:
I think the new parameters logfile and pidfile are also mandatory

hughcapet:
@mblsf as far as I see, `pidfile` is only required for daemonizing. But I think `logfile` is worth using together with `pgqd -q` in order to avoid mixing up pgqd and Patroni output in container logs, thanks for pointing to it.

sdudoladov:
:+1:

hughcapet:
👍

PR #901, opened 2023-08-15
Don't reset defaults in the pgq_ticker.ini

hughcapet:
As we had pgq daemon using the defaults for a long time, let's avoid an unexpected change of behaviour in pgq processing


hughcapet:
👍

PR #794, opened 2022-10-21
PostgreSQL 15 & Ubuntu 22.04

hughcapet:
- Add PostgreSQL 15 support*
- Remove PostgreSQL 9.6
- Use Ubuntu 22.04 as a base image
- Refactor Dockerfile

The following extensions are currently not available for PG15:
- timescaledb
- pllua

PR #908, opened 2023-08-28
Skip relinking files for `PGVERSION`

owenthereal:
https://github.com/zalando/spilo/pull/794 adds support for pg 15 & ubuntu 22.04. But when [PGVERSION](https://github.com/zalando/spilo/blob/master/postgres-appliance/Dockerfile#L2C5-L2C14) is set to non-15 (e.g. `PGVERSION=14`), we got the following error

```
148.2 base.sh: line 284: cd: /usr/share/postgresql/14/contrib/postgis-3.3: No such file or directory
```

This is because the relinking of the file is hard-coded to `15` while it should be the value of `PGVESION`. A real word use case can be referred to [here](https://github.com/hydradatabase/hydra/blob/015107327a061f621fed388f2cc12107df450f8b/Dockerfile.spilo#L25) where `POSTGRES_BASE_VERSION=14`.


owenthereal:
cc: @hughcapet @sdudoladov

hughcapet:
can you please double-check that your problem is solved in the current master branch state?

PR #915, opened 2023-09-05
allow to use application name for upgrade

sylvainOL:
Instead of using replica IP address, allow to use replica name in case IP address is not reliable (like when using a service mesh). 

This trigger is controller by using an environment variable (`USE_APPLICATION_NAME_IN_UPGRADE`) and default behavior is the previous one.

closes zalando/postgres-operator#1629


sylvainOL:
Hello @hughcapet, is there any chances to review it?

Thanks!

sylvainOL:
Hello @jopadi @idanovinda @hughcapet @Jan-M @sdudoladov,

Can you tell me what you think of this PR?

it would help a lot when postgres is used on top of service mesh such as istio

thanks!

Jan-M:
why not query both fields, and then try depending on env var to use the application name or fall back to ip. i am not such a big fan of templating those parts of sql queries.

Jan-M:
But yes, we will keep an eye on this, if this helps in service mesh world we should help you.

sylvainOL:
Hello @Jan-M,

first thanks for quick reply!


> why not query both fields, and then try depending on env var to use the application name or fall back to ip. i am not such a big fan of templating those parts of sql queries.

I'm not (at all) an expert of SQL (and SQL in python) ^_^ so I did the best I could :)

I'm not sure on how to retrieve the values after the requests:

I should propose something like this and it would work?:

```python
        streaming = {a: l for a, l in self.postgresql.query(
            ("SELECT client_addr, application_name, pg_catalog.pg_{0}_{1}_diff(pg_catalog.pg_current_{0}_{1}(),"
             " COALESCE(replay_{1}, '0/0'))::bigint FROM pg_catalog.pg_stat_replication")
            .format(self.postgresql.wal_name, self.postgresql.lsn_name))}

        def ensure_replica_state(member):
            ip = member.conn_kwargs().get('host')
            lag = streaming.get(ip)
            if os.getenv('USE_APPLICATION_NAME_IN_UPGRADE'):
                lag = streaming.get(member.name)
            if lag is None:
                return logger.error('Member %s is not streaming from the primary', member.name)
```

sylvainOL:
hi @Jan-M,

I tried to apply your comment on the PR :)

greyt75:
Another issue we are seeing with istio is that when the COPY TO PROGRAM is executed for rsync it is sending the the istio 127.0.0.6 ip for PRIMARY_IP

postgres-appliance/major_upgrade/inplace_upgrade.py
```diff
@@ -172,13 +172,15 @@ def ensure_replicas_state(self, cluster):
         """
         self.replica_connections = {}
         streaming = {a: l for a, l in self.postgresql.query(
-            ("SELECT client_addr, pg_catalog.pg_{0}_{1}_diff(pg_catalog.pg_current_{0}_{1}(),"
+            ("SELECT client_addr, application_name, pg_catalog.pg_{0}_{1}_diff(pg_catalog.pg_current_{0}_{1}(),"
```
middagj, line 175: This does not work. The result it put in a dictionary, so it expects to unpack two elements, not three. You need to select either `client_addr` or `application_name`.

postgres-appliance/major_upgrade/inplace_upgrade.py
```diff
@@ -172,13 +172,15 @@ def ensure_replicas_state(self, cluster):
...
              " COALESCE(replay_{1}, '0/0'))::bigint FROM pg_catalog.pg_stat_replication")
             .format(self.postgresql.wal_name, self.postgresql.lsn_name))}
 
         def ensure_replica_state(member):
             ip = member.conn_kwargs().get('host')
             lag = streaming.get(ip)
```
middagj, line 181: This should be in the else of the next if.

middagj (changes requested):

alfsch:
Any progress here? Are you needing some help for this issue?

seekermarcel:
since @sylvainOL seems to be unreachable i took over his changes

sylvainOL:
Hello,
sorry I didn't saw the new comments. I'm not using zalando operator anynore so please @seekermarcel take over

sorry for that

PR #916, opened 2023-09-06
Fix tests cleanup script

idanovinda:
* fix AND operation in docker container filter
* fix the ENV var name used in the filter


postgres-appliance/tests/test_spilo.sh
```diff
@@ -12,7 +12,7 @@ readonly TIMEOUT=120
 
 function cleanup() {
     stop_containers
-    docker ps -q --filter "ancestor=${SPILO_IMAGE},name=${PREFIX}" | xargs docker rm -f
+    docker ps -q --filter="ancestor=${SPILO_TEST_IMAGE}" --filter="name=${PREFIX}" | xargs docker rm -f
```
hughcapet, line 15: maybe we add a comment this does not remove the minio container?:)

PR #918, opened 2023-09-07
Setup default image when cleaning up containers

idanovinda:
* add default image to delete when SPILO_TEST_IMAGE var is unset
* extend readme


hughcapet:
👍

idanovinda:
👍

PR #919, opened 2023-09-07
Enable build and push of dev-16 branch

idanovinda:
* update delivery.yaml to build and push when there is push in dev-16 branch


hughcapet:
Maybe, to be on the safe side, we can also comment tagging out here?

hughcapet:
👍

idanovinda:
👍

PR #921, opened 2023-09-12
Multiarch ppc64le support

sumitd2:
*The multiarch image build has been performed successfully (locally) in an Intel Ubuntu 22.04 container.
*A recent version of the Spilo image has been tested on a production ppc64le Openshift cluster.


hughcapet:
we will not be able to test ppc64le internally, which means we can not take responsibility for the image built for every release (i.e. we will not include this now)

gerrith3:
What would you need to be able to support this?  We enable packages like these for Power because we have many customers who run either a package like this directly, or as a dependency of another package.  We would like to help either with the testing directly, helping to set up a CI environment, or any other method that you might find acceptable.

PR #930, opened 2023-10-03
Update /usr/bin/env to python3 (python not found).

rocket357:
A few of the python scripts in postgres-appliance and spilo_cmd contain "#!/usr/bin/env python" and a few contain "#!/usr/bin/env python3".  I noticed cloning (via /scripts/clone_with_wale.py) fails silently, so investigating further showed "/usr/bin/env: ‘python’: No such file or directory" when run with appropriate envdir.

I suspect these should be updated to explicitly select python3, and quick testing shows that resolves the issue (scripts provide usage information instead of the above "No such file or directory" error).

### DIFF
--- a/postgres-appliance/bootstrap/clone_with_basebackup.py
+++ b/postgres-appliance/bootstrap/clone_with_basebackup.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 import argparse
 import logging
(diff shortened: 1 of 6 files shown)


rocket357:
This appears to be a non-issue for automated runs (the clone issue I was hitting was unrelated, and only while manually running the clone_with_wale.py script does the above issue hit) because the scripts themselves are called with python3 (via configure_spilo.py, which is correctly using python3).

It may make sense to merge this if testing the scripts (i.e. clone_with_wale.py) manually is expected, but I suspect that isn't the case?

I'll leave that up to you guys to decide the relevance of this PR.

PR #933, opened 2023-10-05
Refactor tests

hughcapet:
- wait_all_streaming/wait_zero_lag with configurable replica count
- create table with oids using special func, not in the initial schema
- add pg_repack extension to the initial schema
- separate func for creating timescaledb objects
- set WALE_BACKUP_THRESHOLD_PERCENTAGE to a higher value to ensure wal-e replica bootsrap method in test suite 5
- single verify_clone_upgrade func
- adjust comments and formatting
- ensure test suite 4 actually restores latest backup from PG10 and then upgrades


idanovinda:
👍

hughcapet:
👍

PR #943, opened 2023-11-08
Add errorhandling for ulimit

glemsom:
In case the container is not spawned with `SYS_RESOURCE`, setting unlimited size for coredumps will never work.
Since the defaults in the script is to abort on any uncaught errors - I propose we catch the `ulimit` error, and notify the user about the possible missing capability.

This fixes issues like https://github.com/zalando/postgres-operator/issues/2387

### DIFF
--- a/postgres-appliance/runit/patroni/run
+++ b/postgres-appliance/runit/patroni/run
@@ -21,7 +21,6 @@ fi
 if ! ulimit -c unlimited
 then
    echo "Failed to set unlimited size for coredump"
-   echo "Consider adding the SYS_RESOURCE capability"
 fi
 
 # Only small subset of environment variables is allowed. We don't want accidentally disclose sensitive information


postgres-appliance/runit/patroni/run
```diff
@@ -18,7 +18,11 @@ then
 fi
 
 # Enable core dumps
-ulimit -c unlimited
+if ! ulimit -c unlimited
+then
+   echo "Failed to set unlimited size for coredump"
+   echo "Consider adding the SYS_RESOURCE capability"
```
hughcapet, line 24:
```suggestion
```

hughcapet:
I would abstain from adding any specific configuration suggestions to the logging here but let's add the error handling anyway. Thanks.

idanovinda:
👍

hughcapet:
👍

PR #945, opened 2023-11-12
Adjust extensions build/installation process for pg16

hughcapet:
* Build plprofiler from source
* More extensions support pg16: plprofiler, pgl-ddl-deploy
* Small refactoring, as we don't include pg10 anymore: always install/build decoderbufs, pgvector, pg_mon; never install amcheck separately (it is in contrib); no need in $POSTGIS_SUPER_LEGACY.
* Remove pg_tm_aux, plantuner
* Pllua was accidentally not installed for pg15


idanovinda:
👍

hughcapet:
👍

PR #946, opened 2023-11-13
Small tests fix for pg16

hughcapet:
* Wait for zero lag after "with oids" test and timescaledb installation
* Return a test suite id to a tests comment
* Sync formatting with master branch to decrease amount of diff in the further PR


hughcapet:
👍

PR #948, opened 2023-11-18
Remove redundant code

anasanjaria:
In my opinion, it seems that the first line is redundant and can be safely removed.


anasanjaria:
@CyberDem0n It has been opened since ~2w. Could I please receive feedback?

idanovinda:
👍

hughcapet:
👍

PR #949, opened 2023-11-27
Remove amcheck_next installation, removal

hughcapet:
There is no amcheck_next anymore, as we dropped PG10 support.


hughcapet:
👍

PR #951, opened 2023-11-28
add timescaledb version 2.13.0 to PG16

jopadi:
Add new timescaledb version for PG16

https://github.com/timescale/timescaledb/releases/tag/2.13.0


hughcapet:
we also need to start adding timescaledb to the shared_preload_libraries for 16
https://github.com/zalando/spilo/blob/dev-16/postgres-appliance/scripts/spilo_commons.py#L14

(and then you will get tests failing, they will also require little fixes)

jopadi:
👍

PR #962, opened 2024-01-11
No pg_partman update from <5.X during major upgrade

hughcapet:
Due to the noticeable number of backward-incompatible changes in pg_parman 5.0, we better force users to prepare and check everything before actually upgrading the extension. So if performing major upgrade with pg_partman pre 5.X installed, we skip automatic update of it

### DIFF
--- a/postgres-appliance/major_upgrade/pg_upgrade.py
+++ b/postgres-appliance/major_upgrade/pg_upgrade.py
@@ -117,9 +117,12 @@ class _PostgresqlUpgrade(Postgresql):
         for d in self._get_all_databases():
             conn_kwargs['dbname'] = d
             with get_connection_cursor(**conn_kwargs) as cur:
-                cur.execute('SELECT quote_ident(extname) FROM pg_catalog.pg_extension')
-                for extname in cur.fetchall():
-                    query = 'ALTER EXTENSION {0} UPDATE'.format(extname[0])
+                cur.execute('SELECT quote_ident(extname), extversion FROM pg_catalog.pg_extension')
+                for extname, version in cur.fetchall():
+                    # require manual update to 5.X+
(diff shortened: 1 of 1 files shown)


postgres-appliance/major_upgrade/pg_upgrade.py
```diff
@@ -117,9 +117,12 @@ def update_extensions(self):
...
             conn_kwargs['dbname'] = d
             with get_connection_cursor(**conn_kwargs) as cur:
-                cur.execute('SELECT quote_ident(extname) FROM pg_catalog.pg_extension')
-                for extname in cur.fetchall():
-                    query = 'ALTER EXTENSION {0} UPDATE'.format(extname[0])
+                cur.execute('SELECT quote_ident(extname), extversion FROM pg_catalog.pg_extension')
+                for extname, version in cur.fetchall():
+                    # require manual update to 5.X+
+                    if extname == 'pg_partman' and int(version[0]) < 5:
+                        continue
```
macedigital, line 124: We could consider adding a log message here, instead of silently skipping the update. It might be helpful in debugging / informing users.
```suggestion
                        logger.info("Skipping update of '%s' v%s extension in database=%s", extname, version, d)
                        continue
```

hughcapet: I would even make it a warning

macedigital: Agreed :+1:

macedigital (approved):
Looks good to me

macedigital:
:+1:

hughcapet:
👍

macedigital:
:+1:

macedigital:
:+1:

PR #966, opened 2024-01-26
Ensure correct bin_dir for pg params configuration

hughcapet:
Make inplace upgrade code use the bin_dir of the proper version throught the whole process, so that
ConfigHandler.write_postgresql_conf() validates provided params using the proper PG version.

### DIFF
--- a/postgres-appliance/tests/docker-compose.yml
+++ b/postgres-appliance/tests/docker-compose.yml
@@ -45,6 +45,7 @@ services:
                   loop_wait: 2
                   postgresql:
                     parameters:
+                      wal_decode_buffer_size: '521kB'
                       wal_keep_segments: 8
                       jit: 'off'
               postgresql:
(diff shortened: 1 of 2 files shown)


postgres-appliance/tests/docker-compose.yml
```diff
@@ -46,6 +46,7 @@ services:
                   postgresql:
                     parameters:
                       wal_keep_segments: 8
+                      jit: 'off'
```
macedigital, line 49: This is unreleated to the `bin_dir` changes, correct?

hughcapet: it is the parameter that exists in PG11+ and causes 10->N upgrade failure if bin_dir is used incorrectly
However given we are getting rid of 10 (and likely 11) soon, I should probably add other parameters for regression testing in the future.

FxKu:
:+1:

postgres-appliance/tests/docker-compose.yml
macedigital: Understood. Let's keep the parameter, especially given that PG10 is end-of-life by now and not be used much longer.

macedigital (commented):
It took me a couple of passes to get an understanding of the changes, because the broader context of the full code base must be considered.

This the pull-request _should_ fix the failed upgrade from 10 -> 15 which triggered this PR, but I'm honestly not 100% certain all side-effects are caught.

macedigital:
:+1:

macedigital:
:+1:

hughcapet:
👍

PR #967, opened 2024-01-29
Switch to IMDSv2 and boto3

### LINKED ISSUE #963: AWS IMDSv2 Support for environment detection
Spilo is awesome and can [automatically detect the environment it is running in](https://github.com/zalando/spilo/blob/91afafb7cf873239e421be9645b309549ed34c95/postgres-appliance/scripts/configure_spilo.py#L388).

To detect when it is running in AWS, it uses the [Instance Metadata Service](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html). Today, it uses v1 of the API, however AWS is [transitioning to v2](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-metadata-transition-to-version-2.html). Nowadays, new EC2 instance machine images often configured to only support v2. 

Without support for detecting AWS using the v2 API, Spilo set the environment to `PROVIDER_UNSUPPORTED`, and will fail to perform certain bootstrap actions, namely backup (aka clone) restoration.

To fix this, we just need to call [IMDSv2's session token endpoint](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-metadata-v2-how-it-works.html) in the `get_provider` function, and if we get a response, we know the environment is still AWS.

idanovinda:
- retrieve and add token to the curl request headers to replace the usage of IMDSv1 to IMDSv2
- replace boto with boto3 library

### DIFF
--- a/postgres-appliance/scripts/callback_aws.py
+++ b/postgres-appliance/scripts/callback_aws.py
@@ -1,28 +1,12 @@
 #!/usr/bin/env python
 
+from botocore.config import Config
 import boto3
 import logging
 import sys
-import time
 import requests
 
 logger = logging.getLogger(__name__)
-
-
-def retry(func):
-    def wrapped(*args, **kwargs):
-        count = 0
-        while True:
-            try:
-                return func(*args, **kwargs)
-            except boto3.exceptions.Boto3Error as e:
-                if count >= 10 or str(e.error_code) not in ('Throttling', 'RequestLimitExceeded'):
-                    raise
-                logger.info('Throttling AWS API requests...')
-                time.sleep(2 ** count * 0.5)
-                count += 1
-
-    return wrapped
 
 
 def get_instance_metadata():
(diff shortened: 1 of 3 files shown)


postgres-appliance/scripts/callback_aws.py
```diff
@@ -15,7 +15,7 @@ def wrapped(*args, **kwargs):
         while True:
             try:
                 return func(*args, **kwargs)
-            except boto.exception.BotoServerError as e:
+            except boto3.exceptions.Boto3Error as e:
```
hughcapet, line 18: do we need this custom retry at all?
https://boto3.amazonaws.com/v1/documentation/api/latest/guide/error-handling.html#:~:text=%3A%0A%20%20%20%20%20%20%20%20raise%20error-,Note,-The%20Boto3%20standard

https://github.com/boto/botocore/blob/develop/botocore/retries/standard.py

laiminhtrung1997:
Dear all,
If I followed the [best practices](https://aws.github.io/aws-eks-best-practices/security/docs/iam/#restrict-access-to-the-instance-profile-assigned-to-the-worker-node) for security, the spilo container's pod could not access IMDS anymore. So I have to configure the pod to run as hostNetwork so that it can access IMDS.
What do you guys think about this scenario?

PR #979, opened 2024-02-29
fix(clone_with_wale): fix script to work with wal-g master

peterparser:
While working with a wal-g binary built from master branch and spilo, I found out that the newer version of wal-g uses `backup_name` instead of `name` in the output of the command, this change make the script not functional during a restore operation.

The change is pretty straightforward, it's also needed to replace the `backup_name` word to `name` to keep everything together and to support also older versions of wal-g.

### DIFF
--- a/postgres-appliance/bootstrap/clone_with_wale.py
+++ b/postgres-appliance/bootstrap/clone_with_wale.py
@@ -57,9 +57,10 @@ def fix_output(output):
     started = None
     for line in output.decode('utf-8').splitlines():
         if not started:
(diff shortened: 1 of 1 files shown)


gabrielmocan:
No need to work with master anymore as wal-g team just cutted a brand new release.

More on #986

gabrielmocan:
> > No need to work with master anymore as wal-g team just cutted a brand new release.
> > More on #986
> 
> Hi @gabrielmocan, the problem was affecting wal-g master and from what I see in #987 test logs is also affecting this new release.

There's more work to do in this PR, still a draft. Just wait...

PR #984, opened 2024-03-08
Use separate disks for WAL and data

anasanjaria:
One of the performance optimization technique for insertion according to Timescaledb blog is to use separate disks for WAL & data.

According to blog [1],
```
Use separate disks for WAL and data
-----------------------------------
While this is a more advanced optimization that isn't always needed, if your disk becomes a bottleneck,
you can further increase throughput by using a separate disk (tablespace) for the database's write-ahead log (WAL) and data.
```

[1] https://www.timescale.com/blog/13-tips-to-improve-postgresql-insert-performance/


anasanjaria:
@hughcapet Could someone please review my PR as it's opened for a very long time?

vyalamar:
This is a great feature? Can we give this a shot and take it from here?

PR #1006, opened 2024-07-22
Install wal-g from a precompiled bin

hughcapet:
- decrease build time
- fix the problem of missing brotli compression type after upgrade to WAL-G v3 (ref. https://github.com/wal-g/wal-g/pull/1379)

### DIFF
--- a/postgres-appliance/build_scripts/dependencies.sh
+++ b/postgres-appliance/build_scripts/dependencies.sh
@@ -18,36 +18,22 @@ MAKEFLAGS="-j $(grep -c ^processor /proc/cpuinfo)"
 export MAKEFLAGS
 ARCH="$(dpkg --print-architecture)"
 
+# We want to remove all libgdal30 debs except one that is for current architecture.
+printf "shopt -s extglob\nrm /builddeps/!(*_%s.deb)" "$ARCH" | bash -s
+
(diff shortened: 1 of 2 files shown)


macedigital:
Are there tests covering the check whether wal-g supports brotli ? 

It looks like we're relying on the fact that precompiled binaries (may have) enabled the optional brotli compression support by explicitly setting `USE_BROTLI=1` at build time.

hughcapet:
Tests are currently covering our default configuration (not to use WAL-G for backup)
I tested this specific setup though (see the reverted commit)

macedigital:
I think we should add tests for supported compression algorithms in general to Spilo, as `lz4` or `lzma` might be dropped at one time in future as well. This would help detecting issues with creating or restoring from backups early.

Something like: 

```bash
for algo in lz4 lzma brotli; do
  # ... verify backups can be created and restored from $algo
done
```

WDYT?

hughcapet:
depending on how fast you need this fix to be rolled out and if you want to volunteer implementing it now:)

macedigital:
> depending on how fast you need this fix to be rolled out and if you want to volunteer implementing it now:) (this can also be a follow-up PR)

Could be a follow-up PR as fixing applying a fix for the missing brotli support is urgent.

macedigital:
:+1:

hughcapet:
👍

PR #1024, opened 2024-09-04
add extra tags for uploaded Postgres log files

FxKu:
This PR suggest to add some extra tags along with the uploaded log files so that external log analyzer tools (if they are too stupid to parse the import file path) know from which database cluster, K8s namespace and AWS account the log is coming from

The account info is only injected as environment variable by our deployment tool. Therefore, I came up with the idea to make the tags configurable. One can choose keys and values should reference existing env vars.

### DIFF
--- a/postgres-appliance/scripts/upload_pg_log_to_s3.py
+++ b/postgres-appliance/scripts/upload_pg_log_to_s3.py
@@ -64,13 +64,13 @@ def upload_to_s3(local_file_path):
     chunk_size = 52428800  # 50 MiB
     config = TransferConfig(multipart_threshold=chunk_size, multipart_chunksize=chunk_size)
     tags = {
-        'LogEndpoint': os.getenv('LOG_S3_ENDPOINT'),
         'Namespace': os.getenv('POD_NAMESPACE'),
         'ClusterName': os.getenv('SCOPE')
     }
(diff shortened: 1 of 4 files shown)


postgres-appliance/scripts/upload_pg_log_to_s3.py
```diff
@@ -53,9 +53,15 @@ def upload_to_s3(local_file_path):
 
     chunk_size = 52428800  # 50 MiB
     config = TransferConfig(multipart_threshold=chunk_size, multipart_chunksize=chunk_size)
+    tags = eval(os.getenv('LOG_S3_TAGS'))
+    s3_tags = {}
+    for key, value in tags.items():
+        s3_tags[key] = os.getenv(value)
+
+    s3_tags_str = "&".join(f"{key}={value}" for key, value in s3_tags.items())
```
idanovinda, line 61: it seems that we won't use the intermediate variable `s3_tags` anywhere else, would it make sense to directly building the tag string in one step?

```
s3_tags_str = "&".join(f"{key}={os.getenv(value)}" for key, value in tags.items())
```

FxKu: sounds good to me

idanovinda (approved):

hughcapet (approved):

FxKu:
:+1:

PR #1025, opened 2024-09-11
enable hourly log ship mode

FxKu:
The idea is to check in configure_spilo script if the provided schedule contains a / in the second position of the cron schedule. We then set an environment variable to be used in other places to decide on the log file names and Postgre's log_rotation_age.

Open questions:
- does it actually work? Especially the post_init part. Still need to test.
- is there a better way than using an extra env variable?
- more flexibility on different log rotations? For now we decided to only distinguish between daily and houry.

### DIFF
--- a/postgres-appliance/scripts/configure_spilo.py
+++ b/postgres-appliance/scripts/configure_spilo.py
@@ -22,7 +22,7 @@ import yaml
 import pystache
 import requests
 
-from spilo_commons import RW_DIR, PATRONI_CONFIG_FILE, append_extensions,\
+from spilo_commons import RW_DIR, PATRONI_CONFIG_FILE, append_extensions, \
         get_binary_version, get_bin_dir, is_valid_pg_version, write_file, write_patroni_config
 
 
@@ -292,9 +292,19 @@ postgresql:
     logging_collector: 'on'
     log_destination: csvlog
     log_directory: ../pg_log
+    {{#LOG_SHIP_HOURLY}}
     log_filename: 'postgresql-%u.log'
+    {{/LOG_SHIP_HOURLY}}
+    {{^LOG_SHIP_HOURLY}}
+    log_filename: 'postgresql-%u-%H.log'
+    {{/LOG_SHIP_HOURLY}}
     log_file_mode: '0644'
+    {{#LOG_SHIP_HOURLY}}
+    log_rotation_age: '1h'
+    {{/LOG_SHIP_HOURLY}}
+    {{^LOG_SHIP_HOURLY}}
     log_rotation_age: '1d'
+    {{/LOG_SHIP_HOURLY}}
     log_truncate_on_rotation: 'on'
     ssl: 'on'
     {{#SSL_CA_FILE}}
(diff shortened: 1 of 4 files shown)


postgres-appliance/scripts/configure_spilo.py
```diff
@@ -580,6 +590,7 @@ def get_placeholders(provider):
     placeholders.setdefault('CLONE_TARGET_INCLUSIVE', True)
 
     placeholders.setdefault('LOG_GROUP_BY_DATE', False)
+    placeholders.setdefault('LOG_SHIP_HOURLY', False)
```
FxKu, line 593: setting default as false. I would not mention the variable in the ENVIRONMENT.rst file because it has only an internal purpose (could someone still override it?)

FxKu: Changed my mind. It's a lot easier to just mention the flag in a Postgres Operator manifest than coming up with a schedule.

postgres-appliance/tests/test_spilo.sh
```diff
@@ -227,6 +227,19 @@ function start_clone_with_basebackup_upgrade_container() {
...
+    docker-compose run \
+        -e SCOPE=upgrade \
+        -e PGVERSION=16 \
+        -e LOG_SHIP_HOURLY="true" \
+        -e CLONE_SCOPE=demo \
+        -e CLONE_PGVERSION=16 \
+        -e CLONE_METHOD=CLONE_WITH_WALE \
+        -e CLONE_TARGET_TIME="$(next_minute)" \
+        --name "${PREFIX}hourlylogs" \
+        -d "spilo3"
```
hughcapet, line 240:
```suggestion
function start_clone_with_hourly_log_rotation() {
    docker-compose run \
        -e SCOPE=hourlylogs \
        -e PGVERSION=16 \
        -e LOG_SHIP_HOURLY="true" \
        -e CLONE_SCOPE=upgrade2 \
        -e CLONE_PGVERSION=16 \
        -e CLONE_METHOD=CLONE_WITH_WALE \
        -e CLONE_TARGET_TIME="$(next_minute)" \
        --name "${PREFIX}hourlylogs" \
        -d "spilo3"
```

FxKu:
:+1:

postgres-appliance/scripts/post_init.sh
```diff
@@ -147,19 +147,35 @@ if [ "$PGVER" -ge 14 ]; then
...
+        GRANT SELECT ON TABLE public.failed_authentication_${i} TO robot_zmon;"
+    else
+        for h in $(seq 0 23); do
+            echo "CREATE FOREIGN TABLE IF NOT EXISTS public.postgres_log_${i}_${h} () INHERITS (public.postgres_log) SERVER pglog
+            OPTIONS (filename '../pg_log/postgresql-${i}-${h}.csv', format 'csv', header 'false');
+            GRANT SELECT ON public.postgres_log_${i}_${h} TO admin;
+
+            CREATE OR REPLACE VIEW public.failed_authentication_${i}_${h} WITH (security_barrier) AS
+            SELECT *
+              FROM public.postgres_log_${i}_${h}
```
hughcapet, line 172: can we still have per-day views that accumulate data from all the per-hour tables?

FxKu: you mean for failed_authentication and also postgres_log (view that concatenates all foreign tables)?

hughcapet: Yes. To have unified format for both modes

hughcapet:
👍

idanovinda:
👍

PR #1031, opened 2024-10-09
fill s3 tags before copying to log env

FxKu:
When executing the upload command manually inside the pod the tagging works:
`envdir "/run/etc/log.d/env" /scripts/upload_pg_log_to_s3.py`

But seems that crontab only sees the variables which we [copy](https://github.com/zalando/spilo/blob/efd6eb46ee04572dfae41b91ed5629ae66b9c1f8/postgres-appliance/scripts/configure_spilo.py#L798). It will set tags with key found in LOG_S3_TAGS, but values will be `None`.

Therefore, we have to fill out the dict before copying it to the log env.

### DIFF
--- a/postgres-appliance/scripts/upload_pg_log_to_s3.py
+++ b/postgres-appliance/scripts/upload_pg_log_to_s3.py
@@ -3,7 +3,6 @@
 
 import boto3
 import os
-import json
 import logging
 import subprocess
 import sys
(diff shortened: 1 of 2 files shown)


postgres-appliance/scripts/upload_pg_log_to_s3.py
```diff
@@ -63,11 +64,10 @@ def upload_to_s3(local_file_path):
 
     chunk_size = 52428800  # 50 MiB
     config = TransferConfig(multipart_threshold=chunk_size, multipart_chunksize=chunk_size)
-    tags = eval(os.getenv('LOG_S3_TAGS'))
-    s3_tags_str = "&".join(f"{key}={os.getenv(value)}" for key, value in tags.items())
 
     try:
-        bucket.upload_file(local_file_path, key_name, Config=config, ExtraArgs={'Tagging': s3_tags_str})
+        bucket.upload_file(local_file_path, key_name, Config=config,
+                           ExtraArgs={'Tagging': json.loads(os.getenv('LOG_S3_TAGS'))})
```
macedigital, line 70: Is the `json.loads` really necessary here? The transformation in https://github.com/zalando/spilo/pull/1031/files#diff-e4dcf964214d58d733ae9874d4a2ab4b8e9eb99213307e1666af5dad21a326bbR791 should have converted the value to a simple "url-like"-formatted string and no further processing or conversion seems needed:

```suggestion
                           ExtraArgs={'Tagging': os.getenv('LOG_S3_TAGS', '')})
```

idanovinda:
👍

FxKu:
:+1:

PR #1033, opened 2024-10-11
do not append hour on daily uploads

FxKu:
seeing the new daily upload next to previous files I think we better leave the hour out of the name when mode is daily.


FxKu:
:+1:

hughcapet:
👍

PR #1061, opened 2024-12-13
[Patroni4] change no_master and master as argument

idanovinda:
* change no_master to no_leader
* change argument --role master to --role primary

### DIFF
--- a/postgres-appliance/scripts/patroni_wait.sh
+++ b/postgres-appliance/scripts/patroni_wait.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-ROLE=master
+ROLE=primary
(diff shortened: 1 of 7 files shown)


postgres-appliance/scripts/patroni_wait.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-ROLE=master
+ROLE=primary
```
hughcapet, line 3: can you plz also change the comment below?
`Waits for ROLE (master or replica).`

hughcapet:
👍

idanovinda:
👍

PR #1069, opened 2025-01-08
Pass TIMESCALEDB argument as environment variable

idanovinda:
This PR updates the Dockerfile to pass the TIMESCALEDB argument as an environment variable, ensuring it is accessible when using the built image. This allows better flexibility and usability by enabling runtime access to the TIMESCALEDB value for scripts or processes within the container.


hughcapet:
👍

idanovinda:
👍

PR #1079, opened 2025-02-02
Fix timescaledb installation logic

hughcapet:
- properly delete *.so for different setups
- always keep the latest compatible and transition timescaledb ver without specifying in a build arg


idanovinda:
👍

hughcapet:
👍

PR #1087, opened 2025-02-25
retain multiple minor versions of timescaledb

idanovinda:
To ensure consistency between the master and replica nodes, we now retain multiple minor versions of TimescaleDB. This prevents version mismatches that could occur when the master runs an older image while the replica is upgraded to a newer version. By keeping multiple minor versions available, we allow a smoother upgrade path and reduce downtime risks. This change ensures replicas can continue syncing properly without breaking replication.


idanovinda:
👍

hughcapet:
👍

idanovinda:
👍

PR #1091, opened 2025-03-05
Make leader tag value for native AWS configurable

idanovinda:
With the update to Patroni 4, the leader tag value in AWS Native will be set to "primary." This change allows users to configure their preferred value while keeping "master" as the default to maintain consistency with Kubernetes behaviour.


hughcapet:
👍

idanovinda:
👍

PR #1097, opened 2025-03-18
Set proper default KUBERNETES_BOOTSTRAP_LABELS

hughcapet:
to avoid `could not parse kubernetes bootstrap_labels as a JSON: JSONDecodeError('Expecting value: line 1 column 1 (char 0)'), reverting to the default: {}` for an unset env


idanovinda:
👍

hughcapet:
👍

PR #1104, opened 2025-04-08
Correctly format URLs when address is IPv6

mikkeloscar:
If an IPv6 address includes port it must be formatted with surrounding `[]`

This adds logic to optionally add surrounding `[]` if the IP address is IPv6.

### DIFF
--- a/postgres-appliance/scripts/configure_spilo.py
+++ b/postgres-appliance/scripts/configure_spilo.py
@@ -268,7 +268,7 @@ bootstrap:
 scope: &scope '{{SCOPE}}'
 restapi:
   listen: ':{{APIPORT}}'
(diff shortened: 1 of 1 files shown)


postgres-appliance/scripts/configure_spilo.py
```diff
@@ -718,6 +722,12 @@ def get_placeholders(provider):
     return placeholders
 
 
+def format_url(host, port):
+    if ":" in host:
+        return "[" + host + "]" + ":" + port
```
FxKu, line 727: so this does also work with ipv4?

mikkeloscar: Yes, if a host (without port) has `:` in it it indicates IPv6 address and we add the `[]` if not we leave it as is.

mikkeloscar:
Closing as this change doesn't seem to be relevant/needed.

PR #1123, opened 2025-05-21
Fix AWS_EC2_METADATA_SERVICE_ENDPOINT* env handling

hughcapet:
Write AWS_EC2_METADATA_SERVICE_ENDPOINT* env files only when actually needed

### DIFF
--- a/postgres-appliance/scripts/configure_spilo.py
+++ b/postgres-appliance/scripts/configure_spilo.py
@@ -812,13 +812,13 @@ def write_log_environment(placeholders):
         tags = {}
 
     log_env['LOG_S3_TAGS'] = "&".join(f"{key}={os.getenv(value)}" for key, value in tags.items())
-    # support for older boto3 versions: https://github.com/boto/botocore/pull/2600
-    if ('AWS_EC2_METADATA_SERVICE_ENDPOINT' in log_env
-            and not log_env['AWS_EC2_METADATA_SERVICE_ENDPOINT'].endswith('/')):
-        log_env['AWS_EC2_METADATA_SERVICE_ENDPOINT'] += '/'
+    if log_env.get('AWS_EC2_METADATA_SERVICE_ENDPOINT'):
+        # support for older boto3 versions: https://github.com/boto/botocore/pull/2600
+        if not log_env['AWS_EC2_METADATA_SERVICE_ENDPOINT'].endswith('/'):
+            log_env['AWS_EC2_METADATA_SERVICE_ENDPOINT'] += '/'
         write_file(log_env['AWS_EC2_METADATA_SERVICE_ENDPOINT'],
                    os.path.join(log_env['LOG_ENV_DIR'], 'AWS_EC2_METADATA_SERVICE_ENDPOINT'), True)
-    if 'AWS_EC2_METADATA_SERVICE_ENDPOINT_MODE' in log_env:
+    if log_env.get('AWS_EC2_METADATA_SERVICE_ENDPOINT_MODE'):
         write_file(log_env['AWS_EC2_METADATA_SERVICE_ENDPOINT_MODE'],
                    os.path.join(log_env['LOG_ENV_DIR'], 'AWS_EC2_METADATA_SERVICE_ENDPOINT_MODE'), True)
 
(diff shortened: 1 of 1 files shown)


postgres-appliance/scripts/configure_spilo.py
```diff
@@ -516,7 +516,6 @@ def has_dual_stack():
                 if sock:
                     sock.close()
         return False
```
hughcapet, line 518:
```suggestion
        return False
 
```

postgres-appliance/scripts/configure_spilo.py
```diff
@@ -516,7 +516,7 @@ def has_dual_stack():
...
-
+
```
hughcapet, line 519:
```suggestion

```

postgres-appliance/scripts/configure_spilo.py
```diff
@@ -813,8 +812,14 @@ def write_log_environment(placeholders):
 
     log_env['LOG_S3_TAGS'] = "&".join(f"{key}={os.getenv(value)}" for key, value in tags.items())
     # support for older boto3 versions: https://github.com/boto/botocore/pull/2600
-    if not log_env['AWS_EC2_METADATA_SERVICE_ENDPOINT'].endswith('/'):
+    if ('AWS_EC2_METADATA_SERVICE_ENDPOINT' in log_env
+            and not log_env['AWS_EC2_METADATA_SERVICE_ENDPOINT'].endswith('/')):
```
macedigital, line 816: Can we add a check that the environment variable isn't empty either here (and for mode), like so:

```python
and `log_env['AWS_EC2_METADATA_SERVICE_ENDPOINT'] != "" and log_env['AWS_EC2_METADATA_SERVICE_ENDPOINT'].endswith('/')
```

Reasoning: `log_env` will create empty string value for anytime a key was accessed (accidentally). Adding "not empty string" test will prevent cases where other seemingly unrelated changes mutate that dictionary but diff in a PR would show this part of the configure script  https://github.com/zalando/spilo/blob/f3f1b5c9cc44cf6e867c8d9eb0c46ccf427caf7e/postgres-appliance/scripts/configure_spilo.py#L785

hughcapet: there is actually a bigger problem here, AWS_EC2_METADATA_SERVICE_ENDPOINT will not be written when already ends with / 🙈

macedigital: Yes, that too

postgres-appliance/scripts/configure_spilo.py
```diff
@@ -812,9 +812,15 @@ def write_log_environment(placeholders):
...
-    # support for older boto3 versions: https://github.com/boto/botocore/pull/2600
-    if not log_env['AWS_EC2_METADATA_SERVICE_ENDPOINT'].endswith('/'):
-        log_env['AWS_EC2_METADATA_SERVICE_ENDPOINT'] += '/'
+    if log_env.get('AWS_EC2_METADATA_SERVICE_ENDPOINT'):
+        # support for older boto3 versions: https://github.com/boto/botocore/pull/2600
+        if not log_env['AWS_EC2_METADATA_SERVICE_ENDPOINT'].endswith('/'):
+            log_env['AWS_EC2_METADATA_SERVICE_ENDPOINT'] += '/'
+        write_file(log_env['AWS_EC2_METADATA_SERVICE_ENDPOINT'],
+                   os.path.join(log_env['LOG_ENV_DIR'], 'AWS_EC2_METADATA_SERVICE_ENDPOINT'), True)
+    if log_env.get('AWS_EC2_METADATA_SERVICE_ENDPOINT_MODE'):
```
macedigital, line 821: Question: How likely that endpoint mode would be set independent of endpoint environment variable ? (could it be indented?)

hughcapet: these are just the aws sdk's envs. so it is fully on it how it handles their values
what we do here is just ensuring that what is set in the env is also passed to the logs shipping script and can be used by boto (because the env for it is "cleared" and we pass vars via the envdir)

https://github.com/zalando/spilo/blob/0206ea8210934ee9e80565c1a111446da751b87b/postgres-appliance/scripts/configure_spilo.py#L1064-L1074

macedigital, line 821: Yes, I'd be great to understand better if allowing to override both settings independently, introduces another edge case.

macedigital:
:+1:

hughcapet:
👍

PR #1127, opened 2025-05-26
Add pgvector to extension whitelist to allow non-superuser creation

idanovinda:
This change adds the `pgvector` extension to extension whitelist `extwlist`, enabling non-superusers to create the extension without requiring superuser privileges.

### DIFF
--- a/postgres-appliance/scripts/configure_spilo.py
+++ b/postgres-appliance/scripts/configure_spilo.py
@@ -320,7 +320,7 @@ postgresql:
     bg_mon.history_buckets: 120
     pg_stat_statements.track_utility: 'off'
     extwlist.extensions: 'btree_gin,btree_gist,citext,extra_window_functions,first_last_agg,hll,\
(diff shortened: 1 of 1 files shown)


postgres-appliance/scripts/configure_spilo.py
```diff
@@ -320,7 +320,7 @@ def deep_update(a, b):
     bg_mon.history_buckets: 120
     pg_stat_statements.track_utility: 'off'
     extwlist.extensions: 'btree_gin,btree_gist,citext,extra_window_functions,first_last_agg,hll,\
-hstore,hypopg,intarray,ltree,pgcrypto,pgq,pgq_node,pg_trgm,postgres_fdw,roaringbitmap,tablefunc,uuid-ossp'
```
FxKu, line 323: is it really called just vector and not pgvector?

macedigital: I understand your concern, it was "counterintuitive" for me as well. See https://github.com/pgvector/pgvector?tab=readme-ov-file#getting-started

PR #1135, opened 2025-06-18
Fix download url and use ubuntu 22.04 for wal-g

idanovinda:
- download URL has changed for wal-g for arm 
- move to ubuntu 22.04


hughcapet:
👍

PR #1141, opened 2025-07-01
fix wal-g download url and use ubuntu 20.04

idanovinda:
since we downgrade wal-g version from 3.0.7 to 3.0.5 in https://github.com/zalando/spilo/pull/1133, the wal-g download url also need to be fixed


hughcapet:
👍

idanovinda:
👍

PR #1144, opened 2025-07-14
Enable build image for arm64 in pull request

idanovinda:
Enable build image for arm64 in pull request


delivery.yaml
```diff
@@ -30,7 +30,7 @@ pipeline:
       docker buildx create --config /etc/cdp-buildkitd.toml --driver-opt network=host --bootstrap --use
 
       # single platform build for PR images!
```
hughcapet, line 32:
```suggestion
```

PR #1147, opened 2025-07-29
Remove spilo_cmd directory

idanovinda:
This PR removed the `spilo_cmd` directory, which contained a CLI tool that was never adopted in practice and has not been maintained. However, it has been dead code and adds unnecessary complexity to the project. After internal discussion, we agreed that removing it is the best course of action to keep the codebase clean, reduce maintenance burden, and avoid confusion for contributors. This change is self-contained and does not affect any other functionality or components.


hughcapet:
👍

PR #1150, opened 2025-08-06
Fix sorting when retrieve backups

idanovinda:
This PR contains fix sorting for the backups, to print selected fields before sorting.


idanovinda:
👍

PR #1155, opened 2025-10-05
pg_basebackup server compression

hughcapet:
Since pg15, it is possible to compress server-side and reduce the transfer bandwidth


FxKu:
:+1:

hughcapet:
👍

PR #1162, opened 2025-11-05
Allow admin selecting from hypopg views

hughcapet:
with the new public schema setup, users can manipulate but not list hypo indexes


idanovinda:
👍

hughcapet:
👍

PR #1170, opened 2026-01-24
fix: implement container-aware CPU detection

jorsol:
Instead of returning the host's physical CPU count, this now checks Cgroup v1/v2 quotas and process affinity. This ensures the application scales based on available container resources rather than host hardware.

### DIFF
--- a/postgres-appliance/scripts/configure_spilo.py
+++ b/postgres-appliance/scripts/configure_spilo.py
@@ -8,11 +8,11 @@ import re
 import os
 from json.decoder import JSONDecodeError
 
-import psutil
 import socket
 import subprocess
 import sys
 import pwd
+import math
 
 from copy import deepcopy
 from six.moves.urllib_parse import urlparse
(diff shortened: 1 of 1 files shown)


jorsol:
Any comment on this?

hughcapet:
please fix the flake8 issues: https://github.com/zalando/spilo/actions/runs/24730351030/job/72342837925?pr=1170

postgres-appliance/scripts/configure_spilo.py
```diff
@@ -528,6 +529,33 @@ def has_dual_stack():
...
+                    q, p = int(f_q.read()), int(f_p.read())
+                    if q > 0: limit = q / p
+        except Exception:
+            pass
+    # Fallback: Process Affinity or Physical Count
+    if limit is None:
+        try:
+            limit = len(os.sched_getaffinity(0))
+        except (AttributeError, NotImplementedError):
+            limit = os.cpu_count() or 1
```
hughcapet: why do you prefer to switch to `os.cpu_count` from the  `psutil.cpu_count()`

jorsol: Neither of these functions is smart enough to handle container limits (quotas) on their own; it's just a fallback when everything else fails.

Both will return the same value, so I prefer to use the built-in Python function.

jorsol:
> please fix the flake8 issues: https://github.com/zalando/spilo/actions/runs/24730351030/job/72342837925?pr=1170

This should be already fixed.

PR #1172, opened 2026-02-12
Set SPILO_PROVIDER to local by default

hughcapet:
Github actions does not fail token request with an excepted exception and later does not get proper rollback with 

```
requests.exceptions.InvalidHeader: Invalid return character or leading space in header: X-aws-ec2-metadata-token
```

So why not having local as default?


idanovinda:
👍

idanovinda:
👍

hughcapet:
👍

PR #1173, opened 2026-02-12
Change logic for keeping ts minor versions

hughcapet:
keep at least 5 minor versions, but ensure compatibility with the lowest/oldest PG version (where possible)

(copied from dev-18 branch)


idanovinda:
👍

hughcapet:
👍

PR #1181, opened 2026-02-18
Update to wal-g v3.0.8

mikkeloscar:
https://github.com/wal-g/wal-g/releases/tag/v3.0.8

This updates to the latest wal-g version.

It was previously pinned at v3.0.5 because of a bug in v3.0.6 (https://github.com/zalando/spilo/pull/1133) however that bug was fixed upstream.

### DIFF
--- a/postgres-appliance/Dockerfile
+++ b/postgres-appliance/Dockerfile
@@ -18,7 +18,7 @@ FROM $BASE_IMAGE as dependencies-builder
 
 ARG DEMO
 
-ENV WALG_VERSION=v3.0.5
+ENV WALG_VERSION=v3.0.8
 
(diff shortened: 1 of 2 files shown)


hughcapet:
@mikkeloscar can you please change the name of the pkg to the existing one ~(seems they like to change pkg names in wal-g 😅 )~ (my bad, it is just about new ubuntu version)?  https://github.com/wal-g/wal-g/releases/tag/v3.0.8

https://github.com/mikkeloscar/spilo/blob/wal-g-v3.0.8/postgres-appliance/build_scripts/dependencies.sh#L31-L35

idanovinda:
👍

hughcapet:
👍

PR #1182, opened 2026-02-20
Updating wal-g version to 3.0.8

idanovinda:
* wal-g 3.0.8 has released containing the bug fix about s3 versioning problem
* updating to wal-g with ubuntu 22.04


idanovinda:
close as it duplicates

PR #1184, opened 2026-02-22
Fix server version request in basebackup.sh

hughcapet:
use `dbname=postgres` to avoid falling back to a non-existent standby db when no dbname is passed


idanovinda:
👍

hughcapet:
👍

PR #1190, opened 2026-03-01
Ensure CLONE_HOST is IPv6 compatible

mikkeloscar:
I asked AI to find other cases like #1189 and it additionally detected this case which could be IPv6 incompatible.

Fix it similar to #1189


hughcapet:
👍

mikkeloscar:
:+1:

PR #1202, opened 2026-06-04
replace md5 with scram-sha-256

FxKu:
because md5 will be removed in future Postgres releases

### DIFF
--- a/kubernetes/spilo_kubernetes.yaml
+++ b/kubernetes/spilo_kubernetes.yaml
@@ -87,7 +87,7 @@ spec:
           value: | ## https://github.com/zalando/patroni#yaml-configuration
             bootstrap:
               initdb:
-                - auth-host: md5
+                - auth-host: scram-sha-256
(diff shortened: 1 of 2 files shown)


hughcapet:
just in case, you have a plan to MiGrAtE 🙌 🌟  the existing deployments with users' pwds encrypted with md5, right?:)

FxKu:
Yes, clients have to be redeployed, ideally automatically when connections start failing (and pods get replaced).

FxKu:
fyi: On K8s with Postgres Operator we should be mostly safe because scram encryption is already the [default behavior](https://github.com/zalando/postgres-operator/blob/master/pkg/cluster/cluster.go#L140). In other setups we have to set the password_encryption config first. I can check if we can also do it here in Spilo.

Turns out, in EC2 setups this field is also already set to scram and it's only existing passwords and pg_hba.conf which are still set to md5

PR #1205, opened 2026-07-15
Set recommended memory values for Patroni

hughcapet:
https://github.com/patroni/patroni#additional-recommendations-linux-glibc


FxKu:
👍

jopadi:
👍

